PR #14, opened 2016-08-18
Fix spurious image verification failure

wjt:
https://phabricator.endlessm.com/T13004

### DIFF
--- a/gnome-image-installer/pages/install/gis-install-page.c
+++ b/gnome-image-installer/pages/install/gis-install-page.c
@@ -68,7 +68,6 @@ G_DEFINE_QUARK(install-error, gis_install_error);
 #define WID(name) OBJ(GtkWidget*,name)
 
 #define IMAGE_KEYRING "/usr/share/keyrings/eos-image-keyring.gpg"
-#define TRUSTED_KEYID "9E0C1250587A279C"
 
 static gboolean
 gis_install_page_prepare_read (GisPage *page, GError **error)
@@ -341,18 +340,27 @@ gis_install_page_prepare (GisPage *page)
 }
 
 static void
+gis_install_page_verify_failed (GisPage *page,
+                                GError  *error)
+{
+  if (error == NULL)
+    {
+      error = g_error_new (GIS_INSTALL_ERROR, 0, _("Image verification error."));
+    }
+
+  gis_store_set_error (error);
+  g_error_free (error);
+  gis_install_page_teardown(page);
+}
+
+static void
 gis_install_page_gpg_watch (GPid pid, gint status, GisPage *page)
 {
-  GError *error = NULL;
-
   if (!g_spawn_check_exit_status (status, NULL))
-  {
-    error = g_error_new(GIS_INSTALL_ERROR, status, _("Image verification error."));
-    gis_store_set_error (error);
-    g_error_free (error);
-    gis_install_page_teardown(page);
-    return;
-  }
+    {
+      gis_install_page_verify_failed (page, NULL);
+      return;
+    }
 
   gtk_progress_bar_set_fraction (OBJ (GtkProgressBar*, "install_progress"), 0.0);
   g_spawn_close_pid (pid);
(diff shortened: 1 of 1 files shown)


gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -85,8 +85,8 @@ gis_install_page_prepare_read (GisPage *page, GError **error)
   basename = g_file_get_basename(priv->image);
   if (basename == NULL)
     {
-      *error = g_error_new(GIS_INSTALL_ERROR, 0, _("Image verification error: g_file_get_basename(\"%s\") failed"), priv->image);
-      return FALSE;
+      *error = g_error_new (GIS_INSTALL_ERROR, 0, _("Internal error"));
```
sgnn7: Reducing the verbosity of the errors is probably going to cause much harder debugging in user-submitted error reports so I'd leave the originals as-is.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -421,8 +429,55 @@ gis_install_page_verify (GisPage *page)
...
+
+static gboolean
+gis_install_page_prepare_verify (GisPage *page)
+{
+
+  GSubprocess *gpg_import;
+  GError *error = NULL;
+
+  gpg_import = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE, &error,
+      "gpg", "--import", IMAGE_KEYRING, NULL);
```
sgnn7: Seems pretty wrong that we import the whole keyring into the user's gpg keyring (even if this is the `root` user). If you want to go this route, use the `--homedir` param with a tempfile that you clean up afterwards

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -380,38 +388,38 @@ gis_install_page_gpg_progress (GIOChannel *source, GIOCondition cond, GisPage *p
...
-#define ARGS 12
-  gchar *args[ARGS] = { "gpg",
-                        "--enable-progress-filter", "--status-fd", "1",
-                        "--keyring", IMAGE_KEYRING,
-                        "--trusted-key", TRUSTED_KEYID,
-                        "--verify", "", "", NULL };
+  gchar *args[] = { "gpg",
+                    "--enable-progress-filter", "--status-fd", "1",
+                    "--trusted-key", TRUSTED_KEYID,
+                    "--verify", signature_path, image_path, NULL };
```
sgnn7: See below but use a a separate keyring with `--homedir` if you go this route

gnome-image-installer/pages/install/gis-install-page.c
sgnn7: Actually, shouldn't the signing key already be somewhere on the computer? If so, importing is redundant.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -421,8 +429,55 @@ gis_install_page_verify (GisPage *page)
...
+      gint code = g_subprocess_get_exit_status (gpg_import);
+      GError *error = g_error_new (GIS_INSTALL_ERROR, code,
+                                   _("Could not import Endless OS key"));
+      gis_install_page_verify_failed (page, error);
+    }
+
+  g_object_unref (gpg_import);
+}
+
+
```
sgnn7: `dd`

sgnn7:
@wjt Left a couple of comments

gnome-image-installer/pages/install/gis-install-page.c
wjt: These are only "original" as of 24 hours ago when I changed them from just "Image verification error." :smile: 

I'm in two minds on this: I don't want to add strings that are incomprehensible to translators; but you're right that it's easier to debug with more details. I guess I could use strings unmarked for translation for errors which should never occur?

gnome-image-installer/pages/install/gis-install-page.c
wjt: This is the `gnome-initial-setup` user, whose homedir is ephemeral; and this is part of the installer, not a persistent image.

The signing key is already on the system (at `IMAGE_KEYRING`) but if you just pass `--keyring` to GPG, it tries to write a temp file alongside that path, apparently…

If you know a way to do this in one `gpg` invocation… :)

gnome-image-installer/pages/install/gis-install-page.c
sgnn7: Yeah, translations complicate things but at the very least the long form of the error condition should go in some log output somewhere.

wjt:
Updated:
- Single-shot call to `gpg` as discussed
- Verbose messages for programming errors, which will appear in the journal
- Much less code (as a result of the first bullet!)

sgnn7:
@wjt LGTM

PR #17, opened 2016-09-04
T12548 make installer work from live systems

ramcq:
Assuming that the eoslive partition can be mounted by udisks, this allows the installer to detect if it is running in live mode and to install from an uncompressed image stored in endless/endless.img. The endless/eos-eosXX-etc-etc.img.asc signature is used for parsing the image version, personality, etc.

For https://phabricator.endlessm.com/T12548


wjt:
I have updated this PR with a number of cool new features:
1. it applies to `master`
2. it compiles
3. it (probably) works

(In particular, it deals with `endless.img` being hidden from `readdir` and being unreadable even if you find it.)

eos-installer-data/20-eos-installer.rules
```diff
@@ -1,13 +1,23 @@
...
 //
 // DO NOT EDIT THIS FILE, it will be overwritten on update.
 //
 // Allow the gnome-initial-setup user to do certain actions without
 // being interrupted by password dialogs
 
 
+var isLiveSystem;
+try {
+    polkit.spawn(['/bin/grep', '--quiet', 'endless.live_boot', '/proc/cmdline']);
```
ramcq: Possibly excessively picky but, is it worth doing grep -e \bendless.live_boot\b ?

ramcq: (Also endless.live_boot in any case.)

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -242,16 +267,95 @@ static void add_image(GtkListStore *store, gchar *image)
...
   gchar *file = NULL;
   gchar *ufile = NULL;
   GtkListStore *store = OBJ(GtkListStore*, "image_store");
-  GDir *dir = g_dir_open (path, 0, &error);
+  GDir *dir;
   GtkTreeIter iter;
+  gboolean is_live = gis_store_is_live_install ();
+
+  if (is_live)
+    imgpath = g_build_path ("/", path, "endless", NULL);
```
ramcq: I guess this is maybe a little academic, but perhaps to "keep the meat from the flies" when we are acting on separate .img.*z{,.asc} files with the installer running under a live system, rather than reformatting the host with the running live system, maybe we should still look in the root folder and not do this "endless" subdir check any more?

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -273,19 +377,30 @@ gis_diskimage_page_populate_model(GisPage *page, gchar *path)
...
         {
-          add_image(store, fullpath);
+          add_image (store, fullpath, NULL);
         }
       g_free (fullpath);
     }
 
+  if (is_live &&
+      /* we only use endless.img if there are no other images in the folder */
+      !gtk_tree_model_get_iter_first (GTK_TREE_MODEL (store), &iter) &&
```
ramcq: Not sure whether this check adds much at this point, given we look for the image and signature specifically by name and there is no possible .asc ambiguity.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -116,8 +120,17 @@ gis_install_page_prepare_read (GisPage *page, GError **error)
...
+    {
+      priv->decompressed = g_converter_input_stream_new (G_INPUT_STREAM (input),
+                                                         G_CONVERTER (priv->decompressor));
+    }
+  else
+    {
+      priv->decompressed = g_object_ref (input);
+    }
+
   g_object_unref (input);
```
ramcq: I know I wrote this, but - could possibly avoid the ref/unref oddity by just moving the unref up into the if (priv->decompressor) path.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
wjt: Good point, I'll drop it.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
wjt: so you would have the layout be:

```
eoslive/
  eos-pqr.img.xz
  eos-pqr.img.xz.asc
  endless/
    endless.img
    eos-xyz.img.asc
    live  (contains "eos-xyz.img")
```

and you'd be offered both the pqr and xyz images?

gnome-image-installer/pages/install/gis-install-page.c
wjt: huh, I see, can do.

wjt:
I have fixed your various comments and also a couple of other things (see the latest 3 commits). I need to make some corresponding packaging changes. I also am not sure that the `eosimages.mount` stuff is actually needed given that we just use udisks to mount it, do you know the background on this?

PR #18, opened 2016-09-28
T12547

GeorgesStavracas:
https://phabricator.endlessm.com/T12547

### DIFF
--- a/gnome-initial-setup/gis-driver.c
+++ b/gnome-initial-setup/gis-driver.c
@@ -417,7 +417,7 @@ GisDriver *
 gis_driver_new (GisDriverMode mode)
 {
   return g_object_new (GIS_TYPE_DRIVER,
-                       "application-id", "org.gnome.InitialSetup",
+                       "application-id", "com.endlessm.Installer",
                        "mode", mode,
                        NULL);
 }


gnome-initial-setup/gis-driver.c
```diff
@@ -417,7 +461,7 @@ GisDriver *
 gis_driver_new (GisDriverMode mode)
 {
   return g_object_new (GIS_TYPE_DRIVER,
-                       "application-id", "org.gnome.InitialSetup",
+                       "application-id", "org.gnome.Installer",
```
wjt: Surely `com.endlessm.Installer`?

gnome-initial-setup/gis-driver.c
```diff
@@ -80,12 +81,37 @@ struct _GisDriverPrivate {
...
+  gboolean is_live_session;
+
   GisDriverMode mode;
 };
 typedef struct _GisDriverPrivate GisDriverPrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE(GisDriver, gis_driver, GTK_TYPE_APPLICATION)
 
+static gboolean
+running_live_session (void)
```
wjt: Why track this here as well? To skip the language page? Seems to be a no-op at the moment.

wjt (changes requested):

gnome-initial-setup/gis-driver.c
wjt: I wonder if it would be cleaner (and easier to test) to add a `--skip-pages` command-line argument. Actually there's already a magic `/tmp/skip_pages_file` which could be used to do this… not a big deal I suppose.

wjt (approved):

PR #23, opened 2016-10-10
Coexist/conflict with gnome-initial-setup

wjt:
For eosinstaller images, we still want to replace gnome-initial-setup
entirely; but for other images, we want to ship eos-installer alongside
it. To achieve this, we create a second package,
eos-installer-standalone, which serves as the FBE on eosinstaller
images, and simply conflicts with gnome-initial-setup rather than trying
to divert bits of it out of the way.

Open question: should we worry about people who had eos-installer
installed and are dist-upgrading? If so, we'd need to arrange for the
symlink to be removed if someone upgrades eos-installer _without_
installing eos-installer-standalone as well...

https://phabricator.endlessm.com/T12548


ramcq (approved):

wjt:
@ramcq new patch here because of changes in the other PR.

ramcq (approved):

PR #27, opened 2016-10-20
Include icon in package

wjt:
https://phabricator.endlessm.com/T13838


xaiki:
shouldn't this be merged in #26  ?

PR #28, opened 2016-10-27
check_for_live_boot(): return a value on all branches

wjt:
This function returns gboolean; it should probably actually return
something on this error branch.


pwithnall (approved):
Yes!

PR #31, opened 2016-11-16
T13754

GeorgesStavracas:
https://phabricator.endlessm.com/T13754

### DIFF
--- a/gnome-image-installer/pages/install/gis-install-page.c
+++ b/gnome-image-installer/pages/install/gis-install-page.c
@@ -637,6 +637,9 @@ gis_install_page_class_init (GisInstallPageClass *klass)
   GObjectClass *object_class = G_OBJECT_CLASS (klass);
 
   page_class->page_id = PAGE_ID;
+  page_class->hide_forward_button = TRUE;
+  page_class->hide_backward_button = TRUE;
+  page_class->hide_window_controls = TRUE;
   page_class->locale_changed = gis_install_page_locale_changed;
   page_class->shown = gis_install_page_shown;
   object_class->constructed = gis_install_page_constructed;
--- a/gnome-image-installer/pages/install/gis-install-page.c
+++ b/gnome-image-installer/pages/install/gis-install-page.c
@@ -57,6 +57,9 @@ struct _GisInstallPagePrivate {
   guint gpg_watch;
   UDisksClient *client;
   guint pulse_id;
+
+  GtkWidget *warning_dialog;
+  gboolean   writing;
 };
 typedef struct _GisInstallPagePrivate GisInstallPagePrivate;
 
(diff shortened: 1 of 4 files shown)


gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -71,6 +74,58 @@ G_DEFINE_QUARK(install-error, gis_install_error);
...
+                                              "now will leave this system unbootable."));
+
+  /* Abort reformatting */
+  button = gtk_button_new_with_label (_("Stop Reformatting"));
+  gtk_style_context_add_class (gtk_widget_get_style_context (button), "destructive-action");
+  gtk_dialog_add_action_widget (GTK_DIALOG (priv->warning_dialog), button, GTK_RESPONSE_OK);
+
+  gtk_widget_show (button);
+
+  /* Abort reformatting */
```
wjt: This comment doesn't correspond to the button.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -590,19 +665,46 @@ gis_install_page_shown (GisPage *page)
...
+  g_idle_add ((GSourceFunc)gis_install_page_verify, page);
+
+  /* Hide the titlebar widgets */
+  titlebar = gis_assistant_get_titlebar (gis_driver_get_assistant (page->driver));
+  children = gtk_container_get_children (GTK_CONTAINER (titlebar));
 
+  gtk_header_bar_set_show_close_button (GTK_HEADER_BAR (titlebar), FALSE);
+
+  for (l = children; l != NULL ; l = l->next)
+    gtk_widget_hide (l->data);
```
wjt: `update_navigation_buttons` in `gis-assistant.c` already has logic for hiding navigation buttons on the last page. Rather than duplicating it here, perhaps the page can have a vfunc for whether the buttons should be hidden – or a method to call to explicitly hide the buttons?

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -590,19 +665,46 @@ gis_install_page_shown (GisPage *page)
...
+      return;
+    }
+
+  g_idle_add ((GSourceFunc)gis_install_page_verify, page);
+
+  /* Hide the titlebar widgets */
+  titlebar = gis_assistant_get_titlebar (gis_driver_get_assistant (page->driver));
+  children = gtk_container_get_children (GTK_CONTAINER (titlebar));
 
+  gtk_header_bar_set_show_close_button (GTK_HEADER_BAR (titlebar), FALSE);
```
wjt: Similarly, `prepare_main_window` in `gnome-initial-setup/gis-driver.c` has the logic to show the close button and make the window deletable. Maybe it'd be cleaner to put a method onto GisDriver, what do you think?

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -71,6 +74,58 @@ G_DEFINE_QUARK(install-error, gis_install_error);
...
+  gtk_widget_show (button);
+
+  /* Abort reformatting */
+  button = gtk_button_new_with_label (_("Continue Reformatting"));
+  gtk_dialog_add_action_widget (GTK_DIALOG (priv->warning_dialog), button, GTK_RESPONSE_CANCEL);
+
+  gtk_widget_set_can_default (button, TRUE);
+  gtk_widget_show (button);
+
+  /* Lets avoid accidents */
```
wjt: Let's

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -380,6 +439,8 @@ gis_install_page_copy (GisPage *page)
 
   g_thread_yield();
 
+  priv->writing = TRUE;
```
wjt, line 453: There's a potential race here, where you click the Close button, and then the main thread happens to yield, and only then is this thread scheduled, and it runs for long enough to overwrite a little bit of the disk.

Could add a mutex and a GCancellable (say); have the `delete` callback claim the mutex, check `writing`; if it is false, then cancel the GCancellable, and only then release the mutex. Here, you'd claim the mutex, check the cancellable, set `writing` and then release the mutex.

wjt: Or maybe this race is so unlikely as to not matter.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -71,6 +74,58 @@ G_DEFINE_QUARK(install-error, gis_install_error);
...
+
+  /* If we're not writing, it's still safe to quit */
+  if (!priv->writing)
+    return GDK_EVENT_PROPAGATE;
+
+  priv->warning_dialog = gtk_message_dialog_new (GTK_WINDOW (toplevel),
+                                                 GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT,
+                                                 GTK_MESSAGE_WARNING,
+                                                 GTK_BUTTONS_NONE,
+                                                 _("Abort reformatting the disk?"));
```
wjt: I think the verb should be consistent throughout the message / secondary text / button, but at the moment it's "abort" / "cancelling" / "stop". I think "Stop" is the clearest.

wjt (changes requested):

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -590,7 +590,9 @@ gis_install_page_shown (GisPage *page)
 {
   GisInstallPage *install = GIS_INSTALL_PAGE (page);
   GisInstallPagePrivate *priv = gis_install_page_get_instance_private (install);
-  gchar *msg = g_strdup_printf (_("Step %d of %d"), 1, 2);
+  gchar *msg;
+
+  msg = g_strdup_printf (_("Step %d of %d"), 1, 2);
```
wjt: Not a merge blocker but I don't really think this change was necessary.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -433,7 +494,21 @@ gis_install_page_copy (GisPage *page)
...
+
 out:
+
+  /*
+   * If there's a message dialog asking whether the user wants to quit, and
+   * we finished writing, hide and ignore that dialog.
+   */
+  if (priv->warning_dialog)
+    {
+      gtk_dialog_response (GTK_DIALOG (priv->warning_dialog), GTK_RESPONSE_CANCEL);
```
wjt, line 339: Are you sure it's safe to do this from a non-main thread? Shouldn't this go into `gis_install_page_teardown`?

wjt (changes requested):

wjt (approved):
Looks good!

PR #33, opened 2016-12-15
Introduce EosReformatter and an utility to run it

kvahlman:
EosReformatter is a reboot of the verify and write process, wrapped in
a nice package with a bow on top. Should be more efficient too, eventually.

The runner utility will allow easy testing and other image writing
activities in a more automatable fashion than the full installer.

https://phabricator.endlessm.com/T14341

### DIFF
--- a/gnome-image-installer/util/eos-reformatter.c
+++ b/gnome-image-installer/util/eos-reformatter.c
@@ -30,6 +30,7 @@ struct _EosReformatter
 
   gdouble progress;
   gboolean finished;
+  gboolean write_fd_is_external;
   guint sample_update;
 
   guchar *pool;
@@ -167,7 +168,8 @@ eos_reformatter_dispose (GObject *object)
     }
 
   close (reformatter->read_fd);
-  close (reformatter->write_fd);
+  if (!reformatter->write_fd_is_external)
+    close (reformatter->write_fd);
 
   if (reformatter->sample_update != 0)
     {
@@ -248,6 +250,7 @@ eos_reformatter_get_property (GObject    *object,
       g_value_set_string (value, reformatter->device);
       break;
     case PROP_DEVICE_FD:
+      reformatter->write_fd_is_external = TRUE;
       g_value_set_int (value, reformatter->write_fd);
       break;
     case PROP_PROGRESS:
--- a/gnome-image-installer/util/eos-reformatter.c
+++ b/gnome-image-installer/util/eos-reformatter.c
@@ -75,6 +75,7 @@ enum
   PROP_WRITE_SIZE,
   PROP_SIGNATURE,
   PROP_DEVICE,
+  PROP_DEVICE_FD,
   PROP_PROGRESS,
   N_PROPERTIES
 };
@@ -251,6 +252,9 @@ eos_reformatter_get_property (GObject    *object,
     case PROP_DEVICE:
       g_value_set_string (value, reformatter->device);
       break;
+    case PROP_DEVICE_FD:
+      g_value_set_int (value, reformatter->write_fd);
+      break;
     case PROP_PROGRESS:
       g_value_set_double (value, reformatter->progress);
       break;
@@ -284,6 +288,9 @@ eos_reformatter_set_property (GObject      *object,
     case PROP_DEVICE:
       g_free (reformatter->device);
       reformatter->device = g_value_dup_string (value);
+      break;
+    case PROP_DEVICE_FD:
+      reformatter->write_fd = g_value_get_int (value);
       break;
     case PROP_PROGRESS:
       reformatter->progress = g_value_get_double (value);
@@ -335,6 +342,16 @@ eos_reformatter_class_init (EosReformatterClass *klass)
                          "Disk device to write the image to",
                          NULL  /* default value */,
                          G_PARAM_CONSTRUCT_ONLY | G_PARAM_READWRITE);
+
+  _props[PROP_DEVICE_FD] =
+    g_param_spec_int ("device-fd",
+                      "Device file descriptor",
+                      "The file descriptor to write the image to, can be set "
+                      "before starting reformat to use open file descriptor",
+                      G_MININT  /* minimum value */,
+                      G_MAXINT  /* maximum value */,
+                      0  /* default value */,
+                      G_PARAM_READWRITE);
 
   _props[PROP_PROGRESS] =
     g_param_spec_double ("progress",
@@ -647,6 +664,9 @@ eos_reformatter_decompress (EosReformatter *reformatter, GAsyncQueue *freeq, GAs
 static gint
 eos_reformatter_prepare_write (EosReformatter *reformatter)
 {
+  if (reformatter->write_fd > 0)
+    return reformatter->write_fd;
+
   return open(reformatter->device, O_WRONLY | O_SYNC | O_CREAT);
 }
 
--- a/gnome-image-installer/util/eos-reformatter.c
+++ b/gnome-image-installer/util/eos-reformatter.c
@@ -12,6 +12,9 @@
 #include "gduestimator.h"
 #include "eos-reformatter.h"
 
+typedef struct _EosReformatterClass EosReformatterClass;
+typedef struct _EosAlignedBuffer EosAlignedBuffer;
+
 struct _EosReformatter
 {
   GObject parent;
@@ -34,6 +37,7 @@ struct _EosReformatter
   int used_buffers;
   int buffer_size;
   long page_size;
+  EosAlignedBuffer *boot;
 
   GAsyncQueue *free_queue;
   GAsyncQueue *decomp_queue;
@@ -56,13 +60,11 @@ struct _EosReformatter
   GCancellable *cancellable;
 };
 
-typedef struct _EosReformatterClass EosReformatterClass;
 struct _EosReformatterClass
 {
   GObjectClass parent_class;
 };
 
-typedef struct _EosAlignedBuffer EosAlignedBuffer;
 struct _EosAlignedBuffer
 {
   guchar *ptr;
@@ -702,6 +704,26 @@ eos_reformatter_write (EosReformatter *reformatter, GAsyncQueue *freeq, GAsyncQu
   /* EOS */
   if (buf->len == 0)
     {
+
+      /* Write the first buffer last */
+      if (reformatter->boot != NULL)
+        {
+          wb = lseek (reformatter->write_fd, 0, SEEK_SET);
+
+          if (wb == 0)
+            wb = write (reformatter->write_fd, reformatter->boot->ptr, reformatter->boot->len);
+
+          if (wb <= 0)
+            {
+              g_mutex_lock (&reformatter->thread_mutex);
+              reformatter->error = g_error_new (EOS_REFORMATTER_ERROR, 0, _("Internal error"));
+              g_mutex_unlock (&reformatter->thread_mutex);
+            }
+
+          g_async_queue_push (freeq, reformatter->boot);
+          reformatter->boot = NULL;
+        }
+
       syncfs (reformatter->write_fd);
       g_async_queue_push (freeq, buf);
 
@@ -712,6 +734,17 @@ eos_reformatter_write (EosReformatter *reformatter, GAsyncQueue *freeq, GAsyncQu
       g_mutex_unlock (&reformatter->thread_mutex);
 
       return FALSE;
+    }
+
+  /* Grab the first buffer and write zeroes instead so the result does not
+   * look like a bootable system if we abort in the middle
+   */
+  if (reformatter->boot == NULL)
+    {
+      reformatter->boot = buf;
+      buf = eos_reformatter_get_free_buffer (reformatter, freeq);
+      memset (buf->ptr, 0, reformatter->boot->len);
+      buf->len = reformatter->boot->len;
     }
 
   wb = write (reformatter->write_fd, buf->ptr, buf->len);
--- a/gnome-image-installer/util/eos-reformatter.c
+++ b/gnome-image-installer/util/eos-reformatter.c
@@ -367,16 +367,17 @@ eos_reformatter_class_init (EosReformatterClass *klass)
                          G_PARAM_READWRITE);
 
   _signals[SIG_FINISHED] =
-    g_signal_newv ("finished",
+    g_signal_new ("finished",
                    G_TYPE_FROM_CLASS (gobject_class),
                    G_SIGNAL_RUN_LAST | G_SIGNAL_NO_RECURSE | G_SIGNAL_NO_HOOKS,
-                   NULL /* closure */,
+                   0 /* closure */,
                    NULL /* accumulator */,
                    NULL /* accumulator data */,
-                   NULL /* C marshaller */,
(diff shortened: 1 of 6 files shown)


gnome-image-installer/util/run-reformatter.c
```diff
@@ -0,0 +1,140 @@
...
+#include "eos-reformatter.h"
+
+static GMainLoop *loop = NULL;
+static guint64 initial_estimate = 0;
+static gint64 start_usec = 0;
+static int retval = 0;
+
+static gboolean _quit (gpointer data)
+{
+  g_warning (__FUNCTION__);
```
wjt: Why `g_warning`s?

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+  /* A bit awkwardly, exhaust doesn't apply when we create new buffers */
+  if (reformatter->used_buffers < reformatter->total_buffers)
+    {
+      /* No free buffers, but we can create one */
+      gsize offset = reformatter->used_buffers * (reformatter->buffer_size + reformatter->page_size);
+      guchar *unaligned = reformatter->pool + offset;
+      gsize page_size = reformatter->page_size;
+
+      buf = g_new0(EosAlignedBuffer, 1);
+      buf->ptr = (guchar*) (((gintptr) (unaligned + page_size)) & (~(page_size - 1)));
```
wjt, line 441: I really feel like the buffer allocation etc. would be simpler if we just had a pool of `total_buffers` pointers each allocated with:

```c
if (0 != posix_memalign(&ptr, page_size, buffer_size)) { g_error("oh no"); }
```

rather than allocating a pool and then reimplementing aligned allocations from that pool.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+      buf->len = 0;
+      reformatter->used_buffers++;
+    }
+  else
+    {
+      /* If we can't create a new buffer, we need to wait for one */
+      EosAlignedBuffer *buf;
+
+      while (!exhaust && g_async_queue_length (reformatter->free_queue) < half_buffers);
```
wjt: I'm a bit suspicious of this. Would it not be better to factor out the buffer pool and then have two instances, one for the reader and one for the decompressor, rather than having this `exhaust` flag?

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+        {
+          g_thread_yield();
+        }
+
+      do
+        {
+          g_thread_yield();
+          buf = g_async_queue_try_pop (reformatter->free_queue);
```
wjt, line 453: Rather than spinning, could we not use a blocking pop?

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+    {
+      reformatter->error = g_error_new (EOS_REFORMATTER_ERROR, 0, _("Internal error"));
+      g_error_free (error);
+    }
+  else
+    {
+      const gchar *type = NULL;
+
+      /* If we weren't seeded with the target size, use the file size.
+       * This is of course wrong for compressed files */
```
wjt: I don't think we should have logic which is documented to be wrong!

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+
+  /* A free buffer was available, return it */
+  if (buf != NULL)
+    {
+      buf->len = 0;
+      return buf;
+    }
+
+  /* Initialize our pool, if it's not created yet */
+  if (reformatter->pool == NULL)
```
wjt, line 428: What's the benefit in lazily allocating the pool?

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+  info = g_file_query_info (file,
+                            G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE ","
+                            G_FILE_ATTRIBUTE_STANDARD_SIZE,
+                            G_FILE_QUERY_INFO_NONE,
+                            NULL,
+                            &error);
+
+  if (info == NULL)
+    {
+      reformatter->error = g_error_new (EOS_REFORMATTER_ERROR, 0, _("Internal error"));
```
wjt: Wouldn't it be better to propagate the error from `g_file_query_info`, possibly prefixed with some extra information about the context, and have the error page examine the error domain to work out what to tell the user (with "Internal error" being the fallback for unexpected errors)?

wjt: I'm also not that wild about `reformatter->error` being set from potentially many threads without any locking.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+eos_reformatter_read (EosReformatter *reformatter, GAsyncQueue *outq)
+{
+  gboolean ret = TRUE;
+  guint64 len = 0;
+  EosAlignedBuffer *buf = eos_reformatter_get_free_buffer (reformatter, FALSE);
+
+  if (reformatter->error != NULL)
+    {
+      /* Error is set so send EOS and abort */
+      if (buf != NULL)
```
wjt: Surely `buf == NULL` only if an error has occurred, so you can just write a `NULL` (say) to the `outq` and return.

wjt: (I think `NULL` as the sentinel by which errors are signalled to consumers makes more sense than a freshly-allocated, zero-length buffer.)

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+    }
+
+  return buf;
+}
+
+static gint
+eos_reformatter_prepare_read (EosReformatter *reformatter)
+{
+  GFileInfo *info;
+  GFile *file = g_file_new_for_path (reformatter->image);
```
wjt: If you declare `info` and `file` as follows:

```c
g_autoptr(GFileInfo) info = NULL;
g_autoptr(GFile) file = NULL;
```

then they will be automatically unreffed when they fall out of scope.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+{
+  gdouble progress = 0.0;
+  g_object_get (reformatter, "progress", &progress, NULL);
+  return progress;
+}
+
+GError *
+eos_reformatter_get_error (EosReformatter *reformatter)
+{
+  return reformatter->error;
```
wjt: Who owns the returned `GError *`? It should either be `const`, or the API should be rethought.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+      g_async_queue_push (reformatter->free_queue, outbuf);
+    }
+
+  /* This is fine, we just need to cycle to get more input buffers */
+  if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_PARTIAL_INPUT))
+    {
+      res = G_CONVERTER_CONVERTED;
+    }
+
+  if (error != NULL)
```
wjt: In the `PARTIAL_INPUT` case this branch is wrong.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+}
+
+static gboolean
+eos_reformatter_write (EosReformatter *reformatter, GAsyncQueue *inq)
+{
+  gboolean ret = TRUE;
+  EosAlignedBuffer *buf;
+  gsize wb = 0;
+
+  if (reformatter->error != NULL)
```
wjt: Maybe use a `GCancellable` to indicate the case when all the threads can stop. As a bonus this might let us add a Cancel button (of dubious utility of course…).

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,972 @@
...
+
+  reformatter->estimator = gdu_estimator_new (reformatter->write_total_bytes);
+  g_signal_connect (reformatter->estimator, "notify::usec-remaining",
+                    G_CALLBACK (eos_reformatter_update_progress),
+                    reformatter);
+
+  g_warning ("%s: using %d threads", __FUNCTION__, EOS_THREADS);
+
+#if EOS_THREADS == 1
+  reformatter->write_thread = g_thread_new ("read-write", do_reformat, reformatter);
```
wjt: Each thread needs to own a ref to the reformatter, I think.

wjt (changes requested):

gnome-image-installer/util/run-reformatter.c
kvahlman: Just a debugging habit, I'll remove all of those that are not actual warnings.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Since the pool might be large, it would be awkward if the reformatter is created ahead of time and it would immediately gulp a large chunk of memory. Not sure how relevant the use-case is though.

gnome-image-installer/util/eos-reformatter.c
kvahlman: We still need to carry the size information though, since the decompression step produces variable-sized output (unless we make that step complex by always collecting fixed size buffers until passing it on). We could allocate the memory for the buffers with posix_memalign, though I'm not sure if that's better or worse than a large contiguous blob of memory.

gnome-image-installer/util/eos-reformatter.c
kvahlman: We can, if we can guarantee that it gets popped during dispose. Also the code below this appears to assume buf is non-NULL so it's a bugged logic anyway.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Yes, propagating meaningful errors is definitely the intent, the error handling was a bit rushed in to move forward. Hence the fuzzy ownership as well. We'll probably want the error to be delivered either with the finished signal or separate error signal.

gnome-image-installer/util/eos-reformatter.c
kvahlman: The comment predates the type detection, now that we know this indeed should be turned into an error if the file is compressed. Will fix.

gnome-image-installer/util/eos-reformatter.c
kvahlman: I was initially trying to avoid adding NULL-check clutter for the buffers but indeed it is a cleaner solution than allocating new buffers. will fix.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Good catch! Another bug from the error handling additions...

gnome-image-installer/util/eos-reformatter.c
kvahlman: Well we do have a cancel functionality already, GCancellable makes sense to add as the interface for this.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Yeah, the API should be changed to deliver the error through something better than a simple getter.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Right now we (try to) make sure the threads will exit before dispose finishes, so no explicit ref is needed. If we add refs in the threads, a simple g_object_unref() would not trigger shutdown I believe.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Yes, I agree that two queues would be cleaner. This was band-aid to get the scenario of read and decompress threads being separate work properly. I think the original plan was to have just 2 threads so this wouldn't be an issue but worth fixing to use two queues I guess.

gnome-image-installer/util/eos-reformatter.c
kvahlman: Or not, since I now recalled that not only did I want to avoid NULL checks but also g_async_queue_push() doesn't allow NULL to be pushed.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,1088 @@
...
+  guint64 write_unsynced_bytes;
+
+  GThread *read_thread;
+  GThread *decomp_thread;
+  GThread *write_thread;
+
+  GduEstimator *estimator;
+  GConverter *decompressor;
+
+  GMutex thread_mutex;
```
wjt, line 59: I think we need to be clearer which variables are protected by this mutex.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,1088 @@
...
+    }
+  else
+    {
+      /* Plain read, no GPG verification */
+      buf->len = read (reformatter->read_fd, buf->ptr, reformatter->buffer_size);
+    }
+
+  if (len < 0 || buf->len < 0)
+    {
+      g_mutex_lock (&reformatter->thread_mutex);
```
wjt, line 556: This "lock; assign to `error`; unlock" logic is duplicated many times, it should be in a function. Also: every case fails to check whether `error` is non-`NULL`.

gnome-image-installer/util/eos-reformatter.c
wjt: Doing it here seems very suspicious to me, particularly since many threads call `eos_reformatter_get_free_buffer` without any locking.

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,1088 @@
...
+
+  return open(reformatter->device, O_WRONLY | O_SYNC | O_CREAT);
+}
+
+static gboolean
+eos_reformatter_add_sample (gpointer data)
+{
+  EosReformatter *reformatter = EOS_REFORMATTER(data);
+
+  gdu_estimator_add_sample (reformatter->estimator, reformatter->write_completed_bytes);
```
wjt, line 682: `write_completed_bytes` is modified by a different thread so this read should be guarded by the mutex too, no?

gnome-image-installer/util/eos-reformatter.c
```diff
@@ -0,0 +1,1088 @@
...
+}
+
+gboolean
+eos_reformatter_reformat (EosReformatter *reformatter, GCancellable *cancellable)
+{
+  g_return_val_if_fail (reformatter != NULL, FALSE);
+  g_return_val_if_fail (reformatter->image != NULL, FALSE);
+  g_return_val_if_fail (reformatter->device != NULL, FALSE);
+
+  g_object_set (reformatter, "progress", 0.0, NULL);
```
wjt, line 984: We should have some guard against this being called more than once.

timofonic:
@endlessm @wjt  @kvahlman  What happened to this?

wjt:
Superseded by #56.

PR #34, opened 2017-01-02
T14343

kvahlman:
Integrate EosReformatter to the main application, along with improvements to the Unattended mode as aligned in the "Installer unattended mode" document.

The changes ended up overlapping so much there was no point creating separate pull requests.

We are still missing the confirmation dialog and we don't do much heuristic for finding targets but proper defining of target Computers and Images works and the flashing procedure should be faster for regular installs as well.

This is incremental update over #33 and includes the changes made in it.

https://phabricator.endlessm.com/T14342
https://phabricator.endlessm.com/T14343

### DIFF
--- a/gnome-image-installer/gnome-image-installer.c
+++ b/gnome-image-installer/gnome-image-installer.c
@@ -240,7 +240,7 @@ read_keys (const gchar *path)
         {
           if (target_computer (keys, vendor, product))
             {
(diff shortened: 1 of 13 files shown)


timofonic:
@wjt @kvahlman Any news?

wjt:
Superseded by #72.

PR #35, opened 2017-01-19
Enable compiler warnings

wjt:
This covers `gnome-image-installer/`. Some code from `gnome-initial-setup/` is used but I'd rather fix those in upstream g-i-s and pull the changes back down. In a few places there were bugs there exposed by warnings in the installer-specific code, so I've fixed those as needed.

https://phabricator.endlessm.com/T15159

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,7 +132,15 @@ fi
 AC_SUBST(INITIAL_SETUP_CFLAGS)
 AC_SUBST(INITIAL_SETUP_LIBS)
 
-GNOME_COMPILE_WARNINGS([error])
+# -Wno-declaration-after-statement:
+#   Mixed decls are used a lot and are harmless.
+# -Wno-packed:
+#   gpt.h structs are naturally packed, so the compiler warns that the
+#   attribute is unnecessary. I am not quite comfortable removing it even
+#   though I believe the compiler's argument: it has value as documentation.
+AX_COMPILER_FLAGS([WARN_CFLAGS], [WARN_LDFLAGS], [], [],
+                  [-Wno-declaration-after-statement -Wno-packed])
+
 GNOME_MAINTAINER_MODE_DEFINES
 
 AC_SUBST(CFLAGS)
--- a/gnome-image-installer/gnome-image-installer.c
+++ b/gnome-image-installer/gnome-image-installer.c
@@ -32,14 +32,6 @@
 #include <glib/gi18n.h>
 #include <udisks/udisks.h>
 
-#ifdef HAVE_CLUTTER
-#include <clutter-gtk/clutter-gtk.h>
-#endif
-
-#ifdef HAVE_CHEESE
-#include <cheese-gtk.h>
-#endif
-
 #include "pages/language/cc-common-language.h"
 #include "pages/language/gis-language-page.h"
 #include "pages/keyboard/gis-keyboard-page.h"
@@ -58,12 +50,6 @@
 /* main {{{1 */
 
 static gboolean force_new_user_mode;
-static const gchar *system_setup_pages[] = {
-    "account",
-    "display",
-    "endless_eula",
-    "location"
-};
 
 typedef void (*PreparePage) (GisDriver *driver);
 
@@ -360,19 +346,7 @@ gis_page_get_type();
   bind_textdomain_codeset (GETTEXT_PACKAGE, "UTF-8");
   textdomain (GETTEXT_PACKAGE);
 
-#ifdef HAVE_CHEESE
-  cheese_gtk_init (NULL, NULL);
-#endif
-
   gtk_init (&argc, &argv);
-  ev_init ();
-
-#if HAVE_CLUTTER
-  if (gtk_clutter_init (NULL, NULL) != CLUTTER_INIT_SUCCESS) {
-    g_critical ("Clutter-GTK init failed");
-    exit (1);
-  }
-#endif
 
   gis_ensure_login_keyring ("gis");
 
@@ -387,7 +361,6 @@ gis_page_get_type();
 
   g_object_unref (driver);
   g_option_context_free (context);
-  ev_shutdown ();
 
   return status;
 }
--- a/gnome-image-installer/gnome-image-installer.c
+++ b/gnome-image-installer/gnome-image-installer.c
@@ -195,7 +195,7 @@ read_keys (const gchar *path)
 }
 
 static void
-mount_and_read_keys ()
+mount_and_read_keys (void)
 {
   GError *error = NULL;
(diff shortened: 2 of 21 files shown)


configure.ac
```diff
@@ -132,8 +132,17 @@ fi
 AC_SUBST(INITIAL_SETUP_CFLAGS)
 AC_SUBST(INITIAL_SETUP_LIBS)
 
-GNOME_COMPILE_WARNINGS([error])
-GNOME_MAINTAINER_MODE_DEFINES
+# -Wsign-compare:
+#   I was motivated to wire up warnings by hours wasted on checking if a
+#   unsigned value is negative, so -Wsign-compare is non-negotiable.
```
pwithnall: If `-Wsign-compare` is missing from the default upstream `AX_COMPILER_FLAGS` warning set and you believe it has universal value, please submit a patch [upstream](https://github.com/peti/autoconf-archive) and CC me.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -365,14 +361,6 @@ gis_page_get_type();
...
 
   gtk_init (&argc, &argv);
-  ev_init ();
-
-#if HAVE_CLUTTER
-  if (gtk_clutter_init (NULL, NULL) != CLUTTER_INIT_SUCCESS) {
-    g_critical ("Clutter-GTK init failed");
-    exit (1);
-  }
-#endif
```
pwithnall, line 375: Why remove these? Deprecated? Unused? Could do with a comment in the commit message.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -103,7 +103,7 @@ gis_diskimage_page_selection_changed(GtkWidget *combo, GisPage *page)
   gis_store_set_object (GIS_STORE_IMAGE, G_OBJECT (file));
   if (g_str_has_suffix (image, ".gz"))
     {
-      gint64 size = get_gzip_disk_image_size (image);
+      guint64 size = get_gzip_disk_image_size (image);
       if (size <= 0)
```
pwithnall: These `size <= 0` comparisons can be clarified as `size == 0` now (in several places below as well).

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -619,7 +617,7 @@ gis_install_page_verify (GisPage *page)
   const gchar *signature_path = gis_store_get_image_signature ();
   GFile *signature = g_file_new_for_path (signature_path);
   gint outfd;
-  gchar *args[] = { "gpg",
+  const gchar *args[] = { "gpg",
```
pwithnall: This could actually be `const gchar * const args[]`, I think? Because both the array and each value are `const`.

Disclaimer: I haven’t tried this, and my knowledge of C array-vs-`const` semantics is poor at best.

gnome-image-installer/pages/install/gis-install-page.c
```diff
@@ -637,7 +635,7 @@ gis_install_page_verify (GisPage *page)
       goto out;
     }
 
-  if (!g_spawn_async_with_pipes (NULL, args, NULL,
+  if (!g_spawn_async_with_pipes (NULL, (gchar **) args, NULL,
```
pwithnall, line 638: IIRC this cast is the best way to tackle this problem, from the times I’ve tried to come up with a better solution in the past.

pwithnall (changes requested):
As a general comment, I would have done the warning fixes first as separate commits per class of warning; then enabled the compiler warnings as a final commit. Allows for separate review and reversion of related groups of things.

However, that’s just a nitpick and I’m not asking you to rearrange all these commits.

configure.ac
wjt: Actually it's already there; I must have added this before switching from the GNOME one, maybe. I'll remove it.

gnome-image-installer/gnome-image-installer.c
wjt: Unused, inherited from gnome-initial-setup's `main()` I assume. I'll amend.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
wjt: Good catch.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -103,34 +104,26 @@ gis_diskimage_page_selection_changed(GtkWidget *combo, GisPage *page)
...
+      required_size = get_disk_image_size (image);
+    }
+
+  /* This should not happen. Before adding the image to the model, we checked
+   * the filename and validated the GPT header.
+   */
+  if (required_size == 0)
+    {
+      g_warn_if_fail (required_size > 0);
+      required_size = G_MAXINT64;
```
pwithnall: This will now warn if the `image` name doesn’t match any of the suffix checks above. Are you confident that won’t be the case? (If so, maybe add an `else { g_assert_not_reached (); }`?)

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -70,7 +70,8 @@ enum {
     IMAGE_SIZE_BYTES,
     IMAGE_FILE,
     IMAGE_SIGNATURE,
-    ALIGN
+    ALIGN,
+    IMAGE_REQUIRED_SIZE
```
pwithnall, line 74: Nitpick: I’d add a trailing comma here to reduce diff noise in future. (But there’s no need to fix this now; just for future pondering.)

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -250,11 +224,31 @@ add_image (
...
+          get_is_valid_eos_gpt (image))
+        {
+          required_size = get_disk_image_size (image);
+        }
 
-      if ((g_str_has_suffix (image, ".img.gz") && get_gzip_is_valid_eos_gpt (image) == 1)
-       || (g_str_has_suffix (image, ".img.xz") && get_xz_is_valid_eos_gpt (image) == 1)
-       || (g_str_has_suffix (image, ".img") && get_is_valid_eos_gpt (image) == 1)
-       || (image_device != NULL && get_is_valid_eos_gpt (image_device) == 1))
+      if (required_size != 0)
```
pwithnall, line 251: Nice!

pwithnall (approved):
Nice one. I’ll merge shortly.

PR #38, opened 2017-01-24
ISO: Enable Linux reformatter

carlocaione:
https://phabricator.endlessm.com/T14308

### DIFF
--- a/gnome-image-installer/gnome-image-installer.c
+++ b/gnome-image-installer/gnome-image-installer.c
@@ -241,16 +241,19 @@ mount_and_read_keys (void)
 }
 
 static gboolean
-check_for_live_boot (void)
+check_for_live_boot (gchar **uuid)
 {
-  const gchar *force = g_getenv ("EI_FORCE_LIVE_BOOT");
+  const gchar *force = g_getenv ("EI_FORCE_LIVE_BOOT_UUID");
   GError *error = NULL;
   g_autofree gchar *cmdline = NULL;
   gboolean live_boot = FALSE;
+  g_autoptr(GRegex) reg = NULL;
+  g_autoptr(GMatchInfo) info = NULL;
 
   if (force != NULL && *force != '\0')
     {
-      g_print ("EI_FORCE_LIVE_BOOT='%s', set live_boot to TRUE\n", force);
+      g_print ("EI_FORCE_LIVE_BOOT_UUID set to %s\n", force);
+      *uuid = g_strdup (force);
       return TRUE;
     }
 
@@ -264,6 +267,14 @@ check_for_live_boot (void)
   live_boot = g_regex_match_simple ("\\bendless\\.live_boot\\b", cmdline, 0, 0);
 
   g_print ("set live_boot to %u from /proc/cmdline: %s\n", live_boot, cmdline);
+
+  reg = g_regex_new ("UUID=([^\\s]*)", 0, 0, NULL);
+  g_regex_match (reg, cmdline, 0, &info);
+  if (g_match_info_matches (info))
+    {
+      *uuid = g_match_info_fetch (info, 1);
+      g_print ("set UUID to %s\n", *uuid);
+    }
 
   return live_boot;
 }
@@ -330,6 +341,7 @@ main (int argc, char *argv[])
   GisDriver *driver;
   int status;
   GOptionContext *context;
+  gchar *uuid = NULL;
 
 gis_page_get_type();
 
@@ -352,8 +364,11 @@ gis_page_get_type();
 
   gis_ensure_login_keyring ("gis");
 
-  if (check_for_live_boot ())
-    gis_store_enter_live_install ();
+  if (check_for_live_boot (&uuid))
+    {
+      gis_store_enter_live_install ();
+      gis_store_set_image_uuid (uuid);
+    }
 
   mount_and_read_keys ();
 
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -340,6 +340,7 @@ file_exists (
  * various reasons is invisible in directory listings. We can determine the
  * name that the image "should" have by reading /endless/live, and find the
  * corresponding signature file.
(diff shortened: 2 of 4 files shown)


gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -511,18 +517,24 @@ gis_diskimage_page_mount (GisPage *page)
...
 
-      if (!g_str_equal (label, udisks_block_get_id_label (block)))
+      if (is_live && g_str_equal (uuid, udisks_block_get_id_uuid (block)))
+        use_uuid = TRUE;
+
+      if (!g_str_equal (label, udisks_block_get_id_label (block)) && !use_uuid)
         continue;
 
-      if (udisks_block_get_hint_ignore (block))
+      if (udisks_block_get_hint_ignore (block) && !use_uuid)
```
wjt: I think this change would break the exFAT USB case, unfortunately. In that case there are two devices with the UUID from the kernel command line:

1. The real partition (eg `/dev/sdb1`). This partition cannot be used because it is mapped to a device-mapper device; attempts to mount it will fail. This one gets flagged as `UDISKS_IGNORE` in `eos-image-boot-generator`.
2. `/dev/mapper/endless-image-device`, another DM device mapped over the whole image host partition by `eos-image-boot-dm-setup` in `eos-boot-helper`. This one actually works, but probably doesn't exist in the `iso9660` case because only `ntfs` and `exfat` are handled in its copy-pasted logic to get the image file extents. 

So with this change, the ignore flag will be ignored. In the ISO case that will be fine, but in the exfat case it may (or may not, I don't think the ordering is defined) pick the real partition and fail to mount it.

I can see a few possible fixes (others may exist):

1. Drop the logic in `eos-image-boot-dm-setup` which gets the image file extents and punches holes in the mapped image
2. Keep that logic, but add an `iso9660` case which doesn't punch out any holes
3. Keep that logic, but only when the host device is read-write (there's no danger of overwriting your Endless image on a read-only device!)
4. Adjust `dracut/image-boot/eos-image-boot-generator` to not set `UDISKS_IGNORE` in the iso9660 case.

I think I prefer 2. It means we consistently always read the signature etc. from `/dev/mapper/endless-image-device`.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -390,13 +396,13 @@ gis_diskimage_page_add_live_image (
...
-      add_image (store, endless_img_path, live_device_path, live_sig);
+      add_image (store, endless_path, live_device_path, live_sig);
     }
   else
     {
       g_print ("can't find image device %s; will use %s directly\n",
-               live_device_path, endless_img_path);
-      add_image (store, endless_img_path, NULL, live_sig);
+               live_device_path, endless_path);
+      add_image (store, endless_path, NULL, live_sig);
```
wjt: This fallback path doesn't make sense when `endless_path` is `.../endless.squash`, right? The signature won't match, and writing the squash image directly to disk wouldn't boot. I think it would be better to skip this path entirely in that case.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
wjt: You could then make the condition above this:

```c
if (!(
        (is_live && g_str_equal (uuid, udisks_block_get_id_uuid (block)) ||
        g_str_equal (label, udisks_block_get_id_label (block))
   ))
    continue;
```

and drop the `use_uuid` variable.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -240,6 +240,36 @@ mount_and_read_keys (void)
...
+static gchar *
+get_image_uuid (void)
+{
+  gchar *uuid = NULL;
+  GError *error = NULL;
+  g_autofree gchar *cmdline = NULL;
+  GRegex *reg;
+  GMatchInfo *info;
+
+  if (!g_file_get_contents ("/proc/cmdline", &cmdline, NULL, &error))
```
wjt: It's a bit sad to be reading and parsing `/proc/cmdline` twice. Maybe merge this function with `check_for_live_boot` below and as a bonus use the value of `EI_FORCE_LIVE_BOOT` as the UUID if set?

gnome-image-installer/gnome-image-installer.c
```diff
@@ -240,6 +240,36 @@ mount_and_read_keys (void)
...
 }
 
+static gchar *
+get_image_uuid (void)
+{
+  gchar *uuid = NULL;
+  GError *error = NULL;
+  g_autofree gchar *cmdline = NULL;
+  GRegex *reg;
+  GMatchInfo *info;
```
wjt: You could declare these as:

```c
g_autoptr(GRegex) reg = NULL;
g_autoptr(GMatchInfo) info = NULL;
```

if you're into that.

wjt (changes requested):

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
carlocaione: > Keep that logic, but add an iso9660 case which doesn't punch out any holes
 
Was testing this but I get `device-mapper: reload ioctl on endless-image-device failed: Device or resource busy` with a table `0 4415616 linear /dev/disk/by-uuid/2017-01-25-18-23-11-00 0`

Probably because we are already DM-ing a portion of the device in `eos-image-boot-setup`.

carlocaione: > Keep that logic, but only when the host device is read-write (there's no danger of overwriting your Endless image on a read-only device!)

Probably this is not going to work with HD emulation where the partition is still RW. Probably I'll go with:
> Adjust dracut/image-boot/eos-image-boot-generator to not set UDISKS_IGNORE in the iso9660 case.

wjt: We DM the same device twice for USBs and dual-boot, but okay, I can believe that CDs behave differently.

wjt (approved):
LGTM

PR #41, opened 2017-03-08
eos-test-mode: Update eos-updater unit usage

pwithnall:
The unit files in eos-updater were recently changed, so that
eos-updater.service is now a unit in its own right, not an alias for
eos-autoupdater.service. Change the unit file handling to reflect that.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T15977

### DIFF
--- a/gnome-initial-setup/pages/language/eos-test-mode
+++ b/gnome-initial-setup/pages/language/eos-test-mode
@@ -48,5 +48,4 @@ if $extra_active; then
 fi
 
 # Disable the updater for this boot
-systemctl stop eos-updater.timer
-systemctl mask --runtime eos-updater.timer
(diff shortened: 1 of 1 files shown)


wjt:
I don't think there is any need to do this – this copy of the file is not included in any deb built by `eos-installer`, or used…

pwithnall:
> I don't think there is any need to do this – this copy of the file is not included in any deb built by eos-installer, or used…

Right, but I don’t want this incorrect code being cargo-culted in future. Ack this patch (if it’s OK), then delete the file?

wjt (commented):
Okay, let's remove this incorrect (but unused) code and delete it all later.

PR #43, opened 2017-05-19
Use decimal gigabytes

wjt:
This matches what Nautilus and other GNOME applications show for the
size of a file or storage device, by virtue of using the same function.
This also means "GB" is localized into the interface language.

In the error case when the selected drive is too small, we show the size
of the image file in bytes, to be absolutely clear how much space is
needed.

https://phabricator.endlessm.com/T12718

### DIFF
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -352,7 +352,7 @@ add_image (
       if (displayname != NULL)
         {
           goffset size_bytes = g_file_info_get_size (fi);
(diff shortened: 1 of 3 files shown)


ramcq (approved):

ramcq:
Arguably superfluous to use g_format_size_full (.., DEFAULT) vs just g_format_size. But also, equivalent and maybe actually slightly faster. ;)

wjt:
Equivalent and, I thought, slightly clearer. If we later want to show byte counts in all places (as Nautilus' Properties dialog and GNOME Disks's whole disk properties do) we can easily change the flag.

PR #50, opened 2017-08-15
finished: adjust success label for DVD vs USB

wjt:
This includes some refactoring to the way we access UDisks (namely, constructing the proxy object only once, not once every time every page is displayed) which I think is nice to have even without this, and a string fix on the finished page.

Having written this patch, I realise we will need something similar in gnome-initial-setup -- which does not use UDisks at present, and I don't think we should re-add the boot device detection logic there. So perhaps we need to rethink the approach here. Perhaps a symlink from a well-known path to the real boot device so we can determine properties of it – or just touch a flag file to indicate that we're booted from optical media?

https://phabricator.endlessm.com/T18708


gnome-image-installer/gnome-image-installer.c
```diff
@@ -374,6 +375,11 @@ gis_page_get_type();
       gis_store_set_image_uuid (uuid);
     }
 
+  udisks_client = udisks_client_new_sync (NULL, &error);
+  if (udisks_client == NULL)
```
ramcq, line 383: Arguably not a new bug, because nothing was checking before whether UDisksClient was NULL or not... but if udisks_client_new_sync can return NULL, is it worth moving the creation into gis_store, so that every time we attempt to use the UDisks client, we will re-try connecting to the daemon?

ramcq (changes requested):

gnome-image-installer/gnome-image-installer.c
wjt: In the previous incarnation, it's true that 2 out of 3 pages didn't check the return value, but `disktarget` did:

```
  priv->client = udisks_client_new_sync(NULL, &error);
  if (priv->client == NULL)
    {
      g_error("Unable to enumerate disks: %s", error->message);
      g_error_free(error);
    }
```

`g_error()` calls `abort()` so the `g_error_free` seems a bit optimistic ;) This branch does the same. It's true that doing this in `main()` rather than when the 2nd (3rd on `eosinstaller` images) page is shown makes it more likely to fail, given that you can't get past the `diskimage` page if `udisks_client_new_sync()` fails.

I don't think retrying once on each attempt to use the client will help: in the case where the language selection page is skipped, if the `diskimage` page can't use udisks, no images will be found, so it'll be impossible to go forwards; and you can't go backwards; so there's no way to cause a retry.

Two options, not mutually exclusive:

* A retry loop?
* An error dialog rather than an unceremonious `g_error`?

wjt:
Updated with a cmdline parser tweak from https://github.com/endlessm/gnome-initial-setup/pull/194

PR #54, opened 2017-10-04
Verify endless.squash directly

wjt:
From 3.3.x onwards, we include .squash.asc in the ISO image and write
"<image basename>.squash" rather than "<image basename>.img" to the
'live' file. This allows us to verify the squashfs image rather than
decompressing it, verifying the decompressed contents, discarding them,
then decompressing it again to write it to disk.

As ever, this is made more complicated by the image-mapping stuff. In fact,
in the ISO case endless.squash will not be mapped. But when creating
exFAT-formatted USB sticks from an ISO, we would like to be able to copy
endless.squash, rather than decompressing it, to save space on the USB
stick. This is supported by the boot scripts, so, hooray, we get to support
it here too.

Includes a nice cleanup of passing all the facts about each image around.

https://phabricator.endlessm.com/T17203

### DIFF
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -56,13 +56,25 @@ G_DEFINE_TYPE_WITH_PRIVATE (GisDiskImagePage, gis_diskimage_page, GIS_TYPE_PAGE)
 
 G_DEFINE_QUARK(image-error, gis_image_error);
 
-/* A device-mapped copy of endless.img used in image boots.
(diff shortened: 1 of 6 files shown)


wjt:
I find it very unsatisfying that there are so many variations (squashfs or not, exfat or iso, mapped devices or not), and that it's so difficult to test them all without building real images. It would be nice if we could do away with having to care about mapped devices and the filesystem, and read the images from the filesystem in the normal way without a performance hit.

wjt:
Decided against this because the parallel verification work will mean we never see the squashfs contents in eos-installer. I'll resurrect the refactoring in some other branch.

PR #57, opened 2017-11-16
Fully remove egg-list-box submodule, propagate errors from autogen.sh

wjt:
I am still a bit suspicious of this pattern:

    if ! test -f ext/libglnx/README.md; then
        git submodule update --init
    fi

because, since we re-use an existing checkout on Jenkins rather than recreating it, the submodule will never actually be updated. However, this problem also exists for other projects which use this pattern (a quick search of my local checkouts turned up flatpak and ostree) and it strikes me that the best solution is to build in a fresh workspace each time.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 # Run this to generate all the initial makefiles, etc.
+set -ex
 
 srcdir=`dirname $0`
 test -z "$srcdir" && srcdir=.


pwithnall (approved):
Seems reasonable.

pwithnall:
> since we re-use an existing checkout on Jenkins rather than recreating it, the submodule will never actually be updated. However, this problem also exists for other projects which use this pattern (a quick search of my local checkouts turned up flatpak and ostree) and it strikes me that the best solution is to build in a fresh workspace each time.

[gnome-control-center](https://git.gnome.org/browse/gnome-control-center/tree/autogen.sh#n21) does an unconditional:
```
git submodule update --init --recursive
```
which I think is probably the right approach here.

I suspect the `if` is a hangover from when `git submodule update` didn’t have an `--init` option.

PR #58, opened 2017-11-16
Check for GPG in configure, and other build tweaks

wjt:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -139,6 +139,12 @@ fi
 GLIB_TESTS
 LIBGLNX_CONFIGURE
 
+# Both the tests and the application invoke the GPG executable
+AC_ARG_VAR([GPG_PATH], [GPG executable])
+AC_PATH_PROG([GPG_PATH], [gpg])
+AS_IF([test -z "$GPG_PATH"], [AC_MSG_ERROR([GPG not found])])
+AC_DEFINE_UNQUOTED([GPG_PATH], ["$GPG_PATH"], [Path to GPG executable])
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -139,6 +139,13 @@ fi
 GLIB_TESTS
 LIBGLNX_CONFIGURE
 
+# Both the tests and the application invoke the GPG executable
+AC_ARG_VAR([GPG_PATH], [GPG executable])
+AC_PATH_PROG([GPG_PATH], [gpg])
+AS_IF([test -z "$GPG_PATH"], [AC_MSG_ERROR([GPG not found])])
+AC_SUBST([GPG_PATH])
```
pwithnall: Nitpick: The [`AC_ARG_VAR` documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.64/html_node/Setting-Output-Variables.html#index-AC_005fARG_005fVAR-1030) says that it implicitly does an `AC_SUBST` on the variable, so you don’t strictly need this `AC_SUBST` (but it does no harm).

pwithnall (approved):
LGTM with a nitpick which you could or could not fix before pushing.

configure.ac
wjt: Good spot; I read this too but I think I got confused between `AC_SUBST` and `AC_DEFINE`.

pwithnall:
> I got confused between `AC_SUBST` and `AC_DEFINE`.

Don’t we all? :-(

PR #59, opened 2017-11-16
Add missing (build-)dep on gnupg

wjt:
The run-time dependency is redundant with the dependency on
eos-image-keyring but one likes to be explicit. The build-dependency is
newly needed by the test suite.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -38,7 +38,8 @@ Build-Depends: cdbs (>= 0.4.41),
 	       libwebkit2gtk-4.0-dev,
 	       libudisks2-dev,
 	       liblzma-dev,
-	       libattr1-dev
+	       libattr1-dev,
+	       gnupg,
 Standards-Version: 3.9.4
 Homepage: https://endlessm.com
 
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -38,7 +38,8 @@ Build-Depends: cdbs (>= 0.4.41),
 	       libwebkit2gtk-4.0-dev,
 	       libudisks2-dev,
 	       liblzma-dev,
-	       libattr1-dev
+	       libattr1-dev,
+	       gnupg
```
pwithnall: Nitpick: Could add a trailing comma here to avoid diff noise in future.

Also, nitpick: Would be nice to be in alphabetical order.

pwithnall (approved):
LGTM with some nitpicks which you could or could not fix before pushing.

debian/control
wjt: Most of these will go away with https://phabricator.endlessm.com/T17082 which seems a good time to sort them rather than introducing noise now. I'll add the comma though!

PR #60, opened 2017-11-16
tests: skip if memfd_create() fails with ENOSYS

wjt:
Our x86-64 Jenkins builders apparently have a kernel too old for this
syscall. This is rather annoying since (as far as I know) all Endless OS
kernels support this syscall. Oh well.

### DIFF
--- a/tests/test-scribe.c
+++ b/tests/test-scribe.c
@@ -149,13 +149,22 @@ fixture_create_memfd (Fixture *fixture)
   g_autoptr(GOutputStream) output = NULL;
   int fd;
   gboolean ret;
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
 
   fd = memfd_create ("target", MFD_CLOEXEC | MFD_ALLOW_SEALING);
   if (fd < 0)
     {
-      perror ("memfd_create failed");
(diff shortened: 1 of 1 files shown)


tests/test-scribe.c
```diff
@@ -373,6 +387,9 @@ test_error (Fixture       *fixture,
   gboolean ret;
   g_autoptr(GError) error = NULL;
 
+  if (g_test_failed ())
```
pwithnall: That’s kind of icky, but I don’t see an alternative. It’s annoying that `g_test_skip()` in the fixture setup function doesn’t actually skip the test.

Nitpick: Would be good to add a comment like `/* Did fixture setup fail? */`.

pwithnall (approved):
Approved with one nitpick, which you can fix or ignore at your pleasure, before merging.

tests/test-scribe.c
wjt: Added. Yes, I would prefer that the test func were not run if the fixture failed/skipped but the docs suggest this is a feature (emphasis mine):

> Calling this function will not stop the test from running, you need to return from the test function yourself. So you can produce additional diagnostic messages or **even continue running the test**.

PR #61, opened 2017-11-23
install: fix size mismatch error message

wjt:
There was previously no space between the second number and the word
"bytes".

The comment above this block reads:

> Check that we've written the same amount of data as we expected from the
> GPT header. This would only fail if there's something seriously wrong with
> the image builder, the decompressor, or the read/write loop above.

Almost by chance, a bug was introduced into the image builder
simultaneously with introducing this check, such that the image size was
not a multiple of the sector size. We calculate the expected size from
the GPT header (which expresses size in 512-byte sectors) so the actual
size could be up to 511 bytes larger! In this case the extra bytes are
harmless (unless your hard disk happens to be exactly sized so that this
makes a difference), but it's still gratifying to have caught this.

https://phabricator.endlessm.com/T20064


pwithnall (approved):
Borderline, but I can see why you might argue this is more correct.

PR #72, opened 2018-01-23
Rework unattended mode

wjt:
This introduces a new file format at a new path, `unattended.ini`, and various new bits of behaviour. The new format and behaviour are documented in `UNATTENDED.md`, added as part of this branch. The old `install.ini` file is still supported, but only for setting the locale: this is written by the Installer for Windows based on the language it's running in when it creates reformatter USBs.

The new behaviour is to skip as many pages as possible with as little config as possible, but fail hard at the first sign of trouble. In the happiest path where the reformatter is confident that this is the right computer, there is a new 30-second confirmation page to give you a chance to cancel. In all other happy paths, exactly one click is required to proceed.

Outstanding:

* [x] In attended mode, add a hidden keystroke to the last page to write an `unattended.ini` based on the settings you just chose.
* [x] Improve error reporting on the final screen (namely, don't always show "Oops, something is wrong with your Endless OS file.").
* [ ] Cleverer target device matching – I suggest we defer this for now, this branch is already very big

https://phabricator.endlessm.com/T14343

### DIFF
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -508,6 +508,7 @@ static void
 gis_diskimage_page_populate_model (GisPage     *page,
                                    const gchar *path)
 {
+  g_autoptr(GFile) path_file = g_file_new_for_path (path);
   g_autoptr(GError) error = NULL;
   const gchar *file = NULL;
   GisUnattendedConfig *config = gis_store_get_unattended_config ();
@@ -526,6 +527,7 @@ gis_diskimage_page_populate_model (GisPage     *page,
       return;
     }
 
+  gis_store_set_object (GIS_STORE_IMAGE_DIR, G_OBJECT (path_file));
   gtk_list_store_clear(store);
 
   while ((file = g_dir_read_name (dir)))
--- a/gnome-image-installer/pages/finished/gis-finished-page.c
+++ b/gnome-image-installer/pages/finished/gis-finished-page.c
@@ -30,6 +30,7 @@
 #include "finished-resources.h"
 #include "gis-finished-page.h"
 #include "gis-store.h"
+#include "gis-dmi.h"
 
 #include <gtk/gtkx.h>
 #include <glib/gstdio.h>
@@ -43,6 +44,8 @@
 
 struct _GisFinishedPagePrivate {
   gint led_state;
+
+  GtkAccelGroup *accel_group;
 };
 typedef struct _GisFinishedPagePrivate GisFinishedPagePrivate;
 
@@ -212,10 +215,89 @@ gis_finished_page_shown (GisPage *page)
     }
 }
 
+static gboolean
+write_unattended_config (gchar  **backup,
+                         GError **error)
+{
+  GFile *image_file = NULL;
+  GFile *image_dir = NULL;
+  g_autoptr(GFile) unattended_ini_file = NULL;
+  g_autofree gchar *unattended_ini = NULL;
+  const gchar *locale = setlocale (LC_ALL, NULL);
+  g_autofree gchar *image = NULL;
+  UDisksBlock *block = NULL;
+  g_autofree gchar *vendor = NULL;
+  g_autofree gchar *product = NULL;
+  const GError *existing_error = gis_store_get_error ();
+
+  /* Refuse to write a new unattended.ini if installation failed, reminding the
+   * user of the error.
+   */
+  if (existing_error != NULL)
+    {
+      g_propagate_error (error, g_error_copy (existing_error));
+      return FALSE;
+    }
+
+  image_file = G_FILE (gis_store_get_object (GIS_STORE_IMAGE));
+  image = g_file_get_basename (image_file);
+  image_dir = G_FILE (gis_store_get_object (GIS_STORE_IMAGE_DIR));
+  unattended_ini_file = g_file_get_child (image_dir, "unattended.ini");
+  unattended_ini = g_file_get_path (unattended_ini_file);
+  block = UDISKS_BLOCK (gis_store_get_object (GIS_STORE_BLOCK_DEVICE));
+
+  if (!gis_dmi_read_vendor_product (&vendor, &product, error))
+    return FALSE;
+
+  return gis_unattended_config_write (unattended_ini, locale, image,
+                                      udisks_block_get_device (block), vendor,
+                                      product, backup, error);
+}
+
+static void
+write_unattended_config_cb (GisFinishedPage *self)
+{
+  g_autofree gchar *backup = NULL;
+  g_autoptr(GError) error = NULL;
+  GtkMessageType message_type = GTK_MESSAGE_INFO;
+  const gchar *message = NULL;
+  g_autofree gchar *secondary = NULL;
+  GtkWidget *dialog = NULL;
+
+  if (write_unattended_config (&backup, &error))
+    {
+      message = _("Unattended installation configuration file created");
+
+      if (backup != NULL)
+        secondary = g_strdup_printf (_("The previous file was renamed to ‘%s’."),
+                                     backup);
+    }
+  else
+    {
+      message_type = GTK_MESSAGE_ERROR;
+      message = _("Unattended installation configuration file could not be created.");
+      secondary = g_strdup (error->message);
+    }
+
+  dialog = gtk_message_dialog_new (GTK_WINDOW (gtk_widget_get_toplevel (GTK_WIDGET (self))),
+                                   GTK_DIALOG_DESTROY_WITH_PARENT,
+                                   message_type,
+                                   GTK_BUTTONS_OK,
+                                   "%s", message);
+  if (secondary != NULL)
+    gtk_message_dialog_format_secondary_text (GTK_MESSAGE_DIALOG (dialog),
+                                              "%s", secondary);
+
+  gtk_dialog_run (GTK_DIALOG (dialog));
+  gtk_widget_destroy (dialog);
+}
+
 static void
 gis_finished_page_constructed (GObject *object)
 {
   GisFinishedPage *page = GIS_FINISHED_PAGE (object);
+  GisFinishedPagePrivate *priv = gis_finished_page_get_instance_private (page);
+  g_autoptr(GClosure) closure = NULL;
 
   G_OBJECT_CLASS (gis_finished_page_parent_class)->constructed (object);
 
@@ -225,7 +307,23 @@ gis_finished_page_constructed (GObject *object)
 
   g_signal_connect (OBJ (GtkButton *, "restart_button"), "clicked", G_CALLBACK(reboot_cb), page);
 
+  /* Use Ctrl+U to write unattended config */
+  priv->accel_group = gtk_accel_group_new ();
+  closure = g_cclosure_new_swap (G_CALLBACK (write_unattended_config_cb), page, NULL);
+  gtk_accel_group_connect (priv->accel_group, GDK_KEY_u, GDK_CONTROL_MASK, 0, closure);
+
   gtk_widget_show (GTK_WIDGET (page));
+}
+
+static void
+gis_finished_page_dispose (GObject *object)
+{
+  GisFinishedPage *self = GIS_FINISHED_PAGE (object);
+  GisFinishedPagePrivate *priv = gis_finished_page_get_instance_private (self);
+
+  g_clear_object (&priv->accel_group);
+
+  G_OBJECT_CLASS (gis_finished_page_parent_class)->dispose (object);
 }
 
 static void
@@ -253,6 +351,15 @@ gis_finished_page_locale_changed (GisPage *page)
   gtk_label_set_markup (OBJ (GtkLabel *, "support_label"), support_markup);
 }
 
+static GtkAccelGroup *
+gis_finished_page_get_accel_group (GisPage *page)
+{
+  GisFinishedPage *self = GIS_FINISHED_PAGE (page);
+  GisFinishedPagePrivate *priv = gis_finished_page_get_instance_private (self);
+
+  return priv->accel_group;
+}
+
 static void
 gis_finished_page_class_init (GisFinishedPageClass *klass)
 {
@@ -261,8 +368,10 @@ gis_finished_page_class_init (GisFinishedPageClass *klass)
 
   page_class->page_id = PAGE_ID;
   page_class->locale_changed = gis_finished_page_locale_changed;
+  page_class->get_accel_group = gis_finished_page_get_accel_group;
   page_class->shown = gis_finished_page_shown;
   object_class->constructed = gis_finished_page_constructed;
+  object_class->dispose = gis_finished_page_dispose;
 }
 
 static void
--- a/gnome-image-installer/util/gis-store.h
+++ b/gnome-image-installer/util/gis-store.h
@@ -42,7 +42,14 @@ typedef enum {
   /* UDisksClient: global shared UDisks client proxy */
   GIS_STORE_UDISKS_CLIENT,
 
-  /* UDisksDrive: drive hosting partition hosting GIS_STORE_IMAGE */
+  /* GFile: mount point of partition holding GIS_STORE_IMAGE. This is not
+   * always the parent directory of GIS_STORE_IMAGE!
+   */
+  GIS_STORE_IMAGE_DIR,
+
+  /* UDisksDrive: drive hosting partition mounted at GIS_STORE_IMAGE_DIR, or
+   * NULL if it can't be determined
+   */
   GIS_STORE_IMAGE_DRIVE,
 
   GIS_STORE_N_OBJECTS


gnome-image-installer/util/gpt.c
```diff
@@ -152,10 +152,16 @@ uint64_t get_disk_size(struct ptable *pt)
...
- * Checks the GPT for validity
- * returns 1 if the GPT is valid, 0 otherwise
+ * is_eos_gpt_valid:
+ * @size: (out) (optional): location to store the disk size, in bytes, if the
+ *  GPT is valid
+ *
+ * Checks the GPT for validity.
+ *
+ * Returns: 1 if the GPT is valid, 0 otherwise
+ *
```
pwithnall: Nitpick: Unnecessary blank comment line here.

pwithnall (dismissed):
Review of commit “gpt: don't read partition table twice”. 👍

pwithnall (commented):
Review of commit “gnome-initial-setup: add missing header”: 👍.

gnome-image-installer/pages/confirm/Makefile.am
```diff
@@ -0,0 +1,28 @@
+
```
pwithnall: Nitpick: Spurious blank line.

gnome-image-installer/pages/confirm/Makefile.am
```diff
@@ -0,0 +1,28 @@
...
+resource_files = $(shell glib-compile-resources --sourcedir=$(srcdir) --generate-dependencies $(srcdir)/confirm.gresource.xml)
+confirm-resources.c: confirm.gresource.xml $(resource_files)
+	$(AM_V_GEN) $(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir) --generate-source $<
+confirm-resources.h: confirm.gresource.xml $(resource_files)
+	$(AM_V_GEN) $(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir) --generate-header $<
+BUILT_SOURCES += confirm-resources.c confirm-resources.h
+
+libgisconfirm_la_SOURCES =			\
+	gis-confirm-page.c gis-confirm-page.h	\
+	$(BUILT_SOURCES)
```
pwithnall: Should `$(BUILT_SOURCES)` be in `nodist_libgisconfirm_la_SOURCES` instead?

gnome-image-installer/pages/confirm/Makefile.am
```diff
@@ -0,0 +1,28 @@
...
+
+libgisconfirm_la_CFLAGS = $(INITIAL_SETUP_CFLAGS) $(IMAGE_INSTALLER_CFLAGS) -I "$(srcdir)/../../../gnome-initial-setup" -I "$(srcdir)/../.." -I "$(srcdir)/../../util" $(WARN_CFLAGS)
```
pwithnall: Nitpick: Technically the `-I` flags should be in `CPPFLAGS`, but meh.

gnome-image-installer/pages/confirm/gis-confirm-page.h
```diff
@@ -0,0 +1,61 @@
...
+#define GIS_IS_CONFIRM_PAGE_CLASS(klass) (G_TYPE_CHECK_CLASS_TYPE ((klass),  GIS_TYPE_CONFIRM_PAGE))
+#define GIS_CONFIRM_PAGE_GET_CLASS(obj)                 (G_TYPE_INSTANCE_GET_CLASS ((obj),  GIS_TYPE_CONFIRM_PAGE, GisConfirmPageClass))
+
+typedef struct _GisConfirmPage        GisConfirmPage;
+typedef struct _GisConfirmPageClass   GisConfirmPageClass;
+
+struct _GisConfirmPage
+{
+  GisPage parent;
+};
```
pwithnall: Nitpick: Use `G_DECLARE_FINAL_TYPE` and BANG the boilerplate is gone.

(This is just a nitpick, since I assume this has all been cargo-culted, and I can see why you might want to keep it consistent with other pages and port them all at a later date. Also, what you have is not wrong.)

gnome-image-installer/pages/confirm/gis-confirm-page.ui
```diff
@@ -0,0 +1,399 @@
...
+                <property name="top_attach">0</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel" id="vendor_label">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="hexpand">True</property>
+                <property name="label">To Be Filled By OEM</property>
```
pwithnall: Should this be translatable? (‘No’ is a fine answer.)

gnome-image-installer/pages/confirm/gis-confirm-page.ui
```diff
@@ -0,0 +1,399 @@
...
+                <property name="top_attach">1</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkLabel" id="product_label">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="halign">start</property>
+                <property name="hexpand">True</property>
+                <property name="label">To Be Filled By OEM</property>
```
pwithnall: Same here.

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+    GtkButton *cancel_button;
+    GtkButton *continue_button;
+
+    guint countdown_source;
+    guint countdown_remaining;
+};
+typedef struct _GisConfirmPagePrivate GisConfirmPagePrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (GisConfirmPage, gis_confirm_page, GIS_TYPE_PAGE);
+G_DEFINE_QUARK(confirm-error, gis_confirm_error);
```
pwithnall: Nitpick: Missing space before `(`.

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+
+  if (priv->countdown_remaining == 0)
+    {
+      gis_confirm_page_advance (self);
+      return G_SOURCE_REMOVE;
+    }
+  else
+    {
+      g_autofree gchar *countdown_message =
+        g_strdup_printf (_("Installation will commence automatically in %d seconds."),
```
pwithnall: Should this message be using plural support (`g_dngettext()`)?

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+      return G_SOURCE_CONTINUE;
+    }
+}
+
+static void
+gis_confirm_page_start_countdown (GisConfirmPage *self)
+{
+  GisConfirmPagePrivate *priv = gis_confirm_page_get_instance_private (self);
+
+  priv->countdown_remaining = 30;
```
pwithnall: How is this chosen? Might a comment briefly explaining why 30 is the magic number be a good idea?

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+    GtkLabel *warning_label;
+
+    GtkBox *wrong_computer_box;
+
+    GtkLabel *countdown_label;
+    GtkButton *cancel_button;
+    GtkButton *continue_button;
+
+    guint countdown_source;
+    guint countdown_remaining;
```
pwithnall: s/`countdown_remaining`/`countdown_remaining_seconds`/ to make the units more obvious?

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+  gboolean target_computer_specified = TRUE;
+  gboolean target_computer_matches = TRUE;
+
+  if (gis_store_get_error() != NULL)
+    {
+      gis_confirm_page_advance (self);
+      return;
+    }
+
+  /* TODO: update DMI labels */
```
pwithnall: (I assume this `TODO` is what the commit message refers to as being updated in a future commit. 👍 )

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+  GisConfirmPage *self = GIS_CONFIRM_PAGE (page);
+  GisConfirmPagePrivate *priv = gis_confirm_page_get_instance_private (self);
+  GFile *image_file;
+  g_autofree gchar *image_basename = NULL;
+  UDisksClient *udisks_client;
+  UDisksBlock *target_block;
+  UDisksDrive *target_drive = NULL;
+  g_autofree gchar *target_model = NULL;
+  g_autofree gchar *target_size = NULL;
+  /* TODO */
```
pwithnall: `TODO`?

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+static void
+gis_confirm_page_shown (GisPage *page)
+{
+  GisConfirmPage *self = GIS_CONFIRM_PAGE (page);
+  GisConfirmPagePrivate *priv = gis_confirm_page_get_instance_private (self);
+  GFile *image_file;
+  g_autofree gchar *image_basename = NULL;
+  UDisksClient *udisks_client;
+  UDisksBlock *target_block;
+  UDisksDrive *target_drive = NULL;
```
pwithnall: Should `target_drive` be `g_autoptr()`? (`udisks_client_get_drive_for_block()` is transfer full.)

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+}
+
+static void
+gis_confirm_page_cancel_clicked (GtkButton *button,
+                                 GisConfirmPage *self)
+{
+  g_autoptr(GError) error =
+    g_error_new (G_IO_ERROR,
+                 G_IO_ERROR_CANCELLED,
+                 _("Unattended installation was cancelled."));
```
pwithnall: Nitpick: This could be `g_error_new_literal()` (or event `const` and on the stack, although that’s a little ickier).

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -0,0 +1,281 @@
...
+  g_resources_register (confirm_get_resource ());
+
+  gtk_widget_init_template (GTK_WIDGET (self));
+}
+
+void
+gis_prepare_confirm_page (GisDriver *driver)
+{
+  if (gis_store_is_unattended ())
+    gis_driver_add_page (driver,
```
pwithnall: I assume `gis_driver_add_page()` takes ownership of the object you pass it? (I haven’t checked.)

pwithnall (commented):
Review of commit “unattended: add confirmation page”. I haven’t reviewed the UI itself, just the code. Most of my comments are questions rather than bugs. Note that, as always, if I prefix a comment with ‘Nitpick:’ that means I don’t care if it’s addressed or ignored in order for the review to pass.

UNATTENDED.md
```diff
@@ -0,0 +1,74 @@
...
+product=X441SA
+
+[Computer 2]
+vendor=GIGABYTE
+product=GB-BXBT-2807
+```
+
+The correct values for `vendor` and `product` for a given system can be found in `/sys/class/dmi/id/sys_vendor` and `/sys/class/dmi/id/product_name` respectively.
+
+If no computer entries are provided, all computers will be reformatted, but explicit confirmation will be required before the process begins. If computers are specified, using the installer on a different computer will show a warning, and again require explicit confirmation to proceed.
```
pwithnall: Nitpick: Might want to clarify that ‘explicit confirmation’ means no automatic timeout.

UNATTENDED.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Image
+
+If there is only one Endless OS image file on the USB stick, and only one disk present in the system (ignoring the one the installer is running from), no configuration is needed: that image will be written to that disk. Otherwise, add an option group beginning `Image` specifying the `filename` and/or `blockdevice` to use:
+
+```ini
+[Image 1]
+filename=eos-eos3.3-amd64-amd64.180115-104625.en.img.gz
+blockdevice=sd
```
pwithnall: Nitpick: Typically, key names in the G world use hyphens to separate words. Could this be `block-device` instead?

UNATTENDED.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+If specified, `blockdevice` can take one of two forms:
+
+* the base disk type (eg `sd` or `mmcblk`). In this case, eos-installer will list disks matching this name (eg `/dev/sda`, `/dev/sdb`), ignore any which correspond to the USB device the installer is running from, and ignore any where media is not present (eg SD card readers with no card inserted). If that leaves a single candidate, the installer can proceed unattended. Otherwise, the process fails.
```
pwithnall: Nitpick: s/eg/e.g./ or, even better, stop using Latin expressions and just say ‘for example’. (As per §(Latin terms) on https://developer.gnome.org/gdp-style-guide/2.32/grammar.html.en.)

UNATTENDED.md
```diff
@@ -0,0 +1,74 @@
...
+* the full device path (eg `/dev/sda`, `/dev/mmcblk0`).
+
+At present, only writing a single image is supported; including more than one option group beginning `Image` is an error. In future, we may support specifying multiple option groups for dual-disk setups.
```
pwithnall: s/beginning/beginning with/

pwithnall (commented):
Review of commit “unattended: add file format documentation”.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -0,0 +1,175 @@
...
+  G_OBJECT_CLASS (gis_unattended_config_parent_class)->finalize (object);
+}
+
+static void
+gis_unattended_config_class_init (GisUnattendedConfigClass *klass)
+{
+  GObjectClass *object_class = G_OBJECT_CLASS (klass);
+
+  object_class->set_property = gis_unattended_config_set_property;
+  // object_class->get_property = gis_unattended_config_get_property;
```
pwithnall: Nitpick: Could just drop this line.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -0,0 +1,175 @@
...
+  GisUnattendedConfig *self = GIS_UNATTENDED_CONFIG (initable);
+  g_autoptr(GError) local_error = NULL;
+
+  if (self->file_path == NULL)
+    {
+      g_set_error (error,
+                   G_FILE_ERROR,
+                   G_FILE_ERROR_INVAL,
+                   "programmer error: GisUnattendedConfig:file-path must "
+                   "not be NULL");
```
pwithnall: If it’s a programmer error, use `g_return_val_if_fail()` rather than `GError`.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -0,0 +1,175 @@
...
+gis_unattended_config_initable_iface_init (GInitableIface *iface)
+{
+  iface->init = gis_unattended_config_initable_init;
+}
+
+GisUnattendedConfig *
+gis_unattended_config_new (const gchar *file_path,
+                           GError **error)
+{
+  g_return_val_if_fail (file_path != NULL, NULL);
```
pwithnall: Nitpick: If you’re going to put preconditions here (which is a good thing), you might as well also put one in for `error`:
```
g_return_val_if_fail (error == NULL || *error == NULL, NULL);
```

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -0,0 +1,175 @@
...
+      "file path",
+      "Path to unattended.ini file",
+      NULL,
+      G_PARAM_WRITABLE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS);
+
+  g_object_class_install_properties (object_class, N_PROPERTIES, props);
+}
+
+static gboolean
+gis_unattended_config_initable_init (GInitable    *initable,
```
pwithnall: Technically the implementation of `GInitable.init` needs to be idempotent, which means saving the `GError` or success value in your private struct.

tests/Makefile.am
```diff
@@ -95,3 +97,21 @@ test_scribe_LDADD = \
 test_scribe_LDFLAGS = \
 	$(WARN_LDFLAGS) \
 	$(NULL)
+
+test_unattended_config_SOURCES = test-unattended-config.c
+test_unattended_config_CFLAGS = \
+	$(INITIAL_SETUP_CFLAGS) \
+	$(IMAGE_INSTALLER_CFLAGS) \
+	-I $(top_srcdir)/ext/libglnx \
+	-I $(top_srcdir)/gnome-image-installer/util \
```
pwithnall: Nitpick: Technically `-I` `CPPFLAGS` yadda yadda autotools losing the will to care about this kind of thing.

tests/test-unattended-config.c
```diff
@@ -0,0 +1,164 @@
...
+  config = gis_unattended_config_new (empty_ini, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (config);
+}
+
+static void
+test_parse_malformed (void)
+{
+  /* This source file is a perfectly good non-keyfile! */
+  g_autofree gchar *empty_ini =
```
pwithnall: Nitpick: `empty_ini` would seem to not be an appropriate variable name here.

tests/test-unattended-config.c
```diff
@@ -0,0 +1,164 @@
...
+{
+  rm_r (fixture->tmpdir);
+  g_clear_pointer (&fixture->tmpdir, g_free);
+}
+
+static void
+test_parse_empty (void)
+{
+  g_autofree gchar *empty_ini =
+    g_test_build_filename (G_TEST_DIST, "unattended/empty.ini", NULL);
```
pwithnall: Nitpick: Instead of the faff of shipping an empty INI file, I’d probably just touch an empty file in the fixture tmpdir for this test. Choose your poison.

pwithnall (commented):
Review of commit “unattended: introduce GisUnattendedConfig stub”.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -140,28 +140,60 @@ read_sanitized_string (const gchar *filename,
...
+ * but this functionality is now split out into a separate file. If a locale is
+ * specified both install.ini and unattended.ini, the latter takes precedence.
+ */
 static void
-read_keys (const gchar *path)
+read_install_ini (const gchar *path)
 {
-  gchar *ini = g_build_path ("/", path, "install.ini", NULL);
-  GKeyFile *keys = g_key_file_new();
+  g_autofree gchar *ini = g_build_path ("/", path, "unattended.ini", NULL);
```
pwithnall: This actually reads `unattended.ini`, not `install.ini` as the name and documentation suggest.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -140,28 +140,60 @@ read_sanitized_string (const gchar *filename,
...
+static void
+read_unattended_ini (const gchar *path)
+{
+  g_autofree gchar *ini = g_build_path ("/", path, "unattended.ini", NULL);
+  g_autoptr(GisUnattendedConfig) config = NULL;
+  g_autoptr(GError) error = NULL;
+
+  config = gis_unattended_config_new (ini, &error);
+  if (config == NULL &&
+      !g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
```
pwithnall: Nitpick: This condition would look a little clearer if it were `error != NULL && !g_error_matches (…)`.

gnome-image-installer/util/gis-store.c
```diff
@@ -137,14 +136,37 @@ void gis_store_clear_error(void)
...
+gis_store_is_unattended (void)
 {
-  return _unattended;
+  return _config != NULL;
+}
+
+/**
+ * gis_store_get_unattended_config:
+ *
+ * Returns: (transfer none): the unattended config, or %NULL if we are not in
```
pwithnall: Missing `(nullable)`

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -173,3 +173,17 @@ gis_unattended_config_new (const gchar *file_path,
                          "file-path", file_path,
                          NULL);
 }
+
+/**
+ * gis_unattended_config_get_key_file:
+ *
+ * This is a transitional function which we should remove in favour of
+ * structured accessors.
```
pwithnall: Nitpick: Maybe turn this into a `FIXME` for easier grepping in future.

pwithnall (commented):
Review of commit “unattended: instantiate GisUnattendedConfig on startup” 👍.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -118,6 +124,53 @@ gis_unattended_config_class_init (GisUnattendedConfigClass *klass)
...
+
+  value = g_key_file_get_string (key_file, group_name, key, &local_error);
+  if (value == NULL)
+    {
+      if (g_error_matches (local_error, G_KEY_FILE_ERROR,
+                           G_KEY_FILE_ERROR_GROUP_NOT_FOUND) ||
+          g_error_matches (local_error, G_KEY_FILE_ERROR,
+                           G_KEY_FILE_ERROR_KEY_NOT_FOUND))
+        {
+          return TRUE;
```
pwithnall: I suspect you should be setting `*value_out = NULL` in this case.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -118,6 +124,53 @@ gis_unattended_config_class_init (GisUnattendedConfigClass *klass)
...
 
+static gboolean
+key_file_get_optional_string (GKeyFile    *key_file,
+                              const gchar *group_name,
+                              const gchar *key,
+                              gchar      **value_out,
+                              GError     **error)
+{
+  g_autoptr(GError) local_error = NULL;
+  gchar *value;
```
pwithnall: Nitpick: I’d `g_autofree` this and `g_steal_pointer()` below.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -174,6 +227,18 @@ gis_unattended_config_new (const gchar *file_path,
                          NULL);
 }
 
+/**
+ * gis_unattended_config_get_locale:
+ *
+ * Returns: (transfer none): the locale specified in @self, or %NULL if none
```
pwithnall: `(nullable)`

pwithnall (commented):
Review of commit “unattended: read locale from GisUnattendedConfig” 🌵 .

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -123,17 +124,33 @@ gis_confirm_page_shown (GisPage *page)
...
   if (gis_store_get_error() != NULL)
     {
       gis_confirm_page_advance (self);
       return;
     }
 
-  /* TODO: update DMI labels */
+  if (!gis_dmi_read_vendor_product (&vendor, &product, &error))
+    {
+      g_warning ("failed to read DMI data: %s", error->message);
```
pwithnall: Nitpick: s/failed/Failed/

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -123,17 +124,33 @@ gis_confirm_page_shown (GisPage *page)
...
+      g_clear_error (&error);
+    }
+
+  if (vendor != NULL)
+    gtk_label_set_text (priv->vendor_label, vendor);
+  else
+    gtk_label_set_markup (priv->vendor_label, _("<i>Unknown</i>"));
```
pwithnall: Nitpick: Ideally, the markup would be applied outside of the string translation, so that translators don’t have to be trusted to preserve the markup.

gnome-image-installer/pages/confirm/gis-confirm-page.c
```diff
@@ -155,18 +172,26 @@ gis_confirm_page_shown (GisPage *page)
 
   g_clear_object (&target_drive);
 
+  config = gis_store_get_unattended_config ();
+  g_assert_nonnull (config);
```
pwithnall: Nerdery: you should use `g_assert()` in application code, and `g_assert_*()` in unit test code, since the former is disabled when compiling with `G_DISABLE_ASSERT` (or whatever it’s called); and the latter is not.

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+#include <gio/gio.h>
+
+#define SYS_VENDOR "/sys/class/dmi/id/sys_vendor"
+#define SYS_PRODUCT "/sys/class/dmi/id/product_name"
+
+/**
+ * gis_dmi_sanitize_string:
+ * @string: (transfer none): an arbitrary bytestring
+ *
+ * Returns: (transfer full): a copy of @string, with bytes outside the ASCII
```
pwithnall: `(nullable)`

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+#include "config.h"
+
+#include <gio/gio.h>
+
+#define SYS_VENDOR "/sys/class/dmi/id/sys_vendor"
+#define SYS_PRODUCT "/sys/class/dmi/id/product_name"
+
+/**
+ * gis_dmi_sanitize_string:
+ * @string: (transfer none): an arbitrary bytestring
```
pwithnall: Presumably with no embedded nuls? I’d lean towards typing it as `const guint8 *` to make it clearer that it’s not UTF-8. I kind of always assume that `const gchar *` is UTF-8.

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+  g_autofree gchar *sanitized = g_strdup (string);
+  const gchar *r = string;
+  gchar *w = sanitized;
+
+  if (string == NULL)
+    return NULL;
+
+  for (; *r != '\0'; r++)
+    {
+      if (*r < 32 || *r > 126)
```
pwithnall: Nitpick: Might be clearer to use character literals for these?

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+      gchar *sanitized = gis_dmi_sanitize_string (contents);
+      if (sanitized == NULL)
+        g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                     "%s contained no printable characters",
+                     filename);
+      return sanitized;
+    }
+  else
+    {
+      g_prefix_error (error, "failed to read %s", filename);
```
pwithnall: Nitpick: s/failed/Failed/

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+      return NULL;
+    }
+}
+
+/**
+ * gis_dmi_read_vendor_product:
+ * @vendor: (out): location to store the computer's vendor
+ * @product: (out): location to store the computer's product name
```
pwithnall: Missing `(transfer full)` on these. Might also want to add `(not optional)`.

gnome-image-installer/util/gis-dmi.c
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * Reads sanitized vendor and product information from DMI tables. The returned
+ * strings are guaranteed to contain only printable ASCII characters, with no
+ * leading or trailing whitespace.
```
pwithnall: Probaby should also mention that, on success, they will both be non-empty strings too.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -162,12 +175,42 @@ static gboolean
 gis_unattended_config_populate_fields (GisUnattendedConfig *self,
                                        GError **error)
 {
+  g_auto(GStrv) groups = g_key_file_get_groups (self->key_file, NULL);
+  gchar **group;
```
pwithnall: `const gchar **`?

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -239,6 +282,46 @@ gis_unattended_config_get_locale (GisUnattendedConfig *self)
...
+ * gis_unattended_config_match_computer:
+ * @vendor: (nullable): the current computer's vendor, or %NULL if it could not
+ *  be determined
+ * @product: (nullable): the current computer's model, or %NULL if it could not
+ *  be determined
+ *
+ * Compares @vendor and @product to the computer(s) listed in the unattended
+ * configuration, ASCII case-insensitively. In the special case where @vendor
+ * or @product is %NULL, this function will never return
+ * GIS_UNATTENDED_COMPUTER_MATCHES.
```
pwithnall: Nitpick: Missing the `%` formatting from the error code.

tests/Makefile.am
```diff
@@ -117,3 +120,19 @@ test_unattended_config_LDADD = \
 test_unattended_config_LDFLAGS = \
 	$(WARN_LDFLAGS) \
 	$(NULL)
+
+test_dmi_SOURCES = test-dmi.c
+test_dmi_CFLAGS = \
+	$(INITIAL_SETUP_CFLAGS) \
+	$(IMAGE_INSTALLER_CFLAGS) \
+	-I $(top_srcdir)/gnome-image-installer/util \
```
pwithnall: Nitpick: `-I` `CPPFLAGS` blah yakkety blah meh whatever.

pwithnall: I shall die on this hill.

Sad and alone.

pwithnall: But ultimately correct.

pwithnall (commented):
Review of commit “unattended: allow matching 0, 1, or more computers”, 🚜. I suggest you change the commit message s/0, 1, or more/Z^≥/.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -135,7 +135,10 @@ read_unattended_ini (const gchar *path)
   if (config == NULL &&
       !g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
     {
-      g_warning ("Error loading %s: %s", ini, error->message);
+      /* Translators: the placeholder is the name of a configuration file, such
+       * as "/run/mount/eosimages/unattended.ini". */
+      g_prefix_error (&error, _("Error loading %s: "), ini);
```
pwithnall: Perhaps this message should quote the filename? “Error loading ‘%s’: ”?

pwithnall (commented):
Review of commit “unattended: enter error state if parsing fails”, 🍰.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -124,13 +124,15 @@ gis_diskimage_page_selection_changed(GtkWidget *combo, GisPage *page)
...
   gis_page_set_complete (page, TRUE);
 
-  if (gis_store_is_unattended())
+  if (gis_store_is_unattended ())
     {
       if (gtk_tree_model_iter_n_children (model, NULL) > 1)
         {
-          GError *error = g_error_new (GIS_IMAGE_ERROR, 0, _("No suitable images were found."));
+          g_autoptr(GError) error =
+            g_error_new (GIS_UNATTENDED_ERROR,
```
pwithnall: Nitpick: This could be `g_error_new_literal()`.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -473,9 +475,10 @@ gis_diskimage_page_add_live_image (
 
   if (ufile != NULL && g_strcmp0 (ufile, live_flag_contents) != 0)
     {
-      g_set_error (error, GIS_IMAGE_ERROR, 0,
-          "live image '%s' doesn't match unattended image '%s'",
-          live_flag_contents, ufile);
+      g_set_error (error, GIS_UNATTENDED_ERROR,
+                   GIS_UNATTENDED_ERROR_IMAGE_NOT_FOUND,
+                   "live image '%s' doesn't match configured image '%s'",
```
pwithnall: Nitpick: s/live/Live/, and s/'/‘/ and s/'/’/

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -209,6 +220,45 @@ gis_unattended_config_populate_fields (GisUnattendedConfig *self,
...
+                                             *group, FILENAME_KEY,
+                                             &self->filename,
+                                             error) ||
+              !key_file_get_optional_string (self->key_file,
+                                             *group, BLOCKDEVICE_KEY,
+                                             &self->blockdevice,
+                                             error))
+            return FALSE;
+
+          if (self->blockdevice != NULL && *self->blockdevice == '\0')
```
pwithnall: Might make sense to factor this validation out into a helper function.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -282,6 +332,44 @@ gis_unattended_config_get_locale (GisUnattendedConfig *self)
...
+{
+  const gchar *basename;
+
+  if (self->blockdevice == NULL)
+    return TRUE;
+
+  if (self->blockdevice[0] == '/')
+    return g_strcmp0 (device, self->blockdevice) == 0;
+
+  basename = rindex (device, '/');
```
pwithnall: Nitpick: It’s not really the basename if you get a pointer to the final `/`, it’s the basename plus a leading `/`.

Why not use `g_path_get_basename()`?

pwithnall (commented):
Review of commit “unattended: match image filename and device path”, 🐤.

pwithnall (commented):
Review for “diskimage: set GIS_STORE_IMAGE_DRIVE sooner”, 🍸. I’m intrigued as to why it was ever set after advancing the page. 🤔

pwithnall (commented):
Review of “run-mount-eosimages.mount: match mount point path”, 👍.

pwithnall (commented):
Review of commit “unattended: remove keyfile accessor”, 👍.

pwithnall (commented):
Review of commit “diskimage: fix some consts and leaks”, 👍. I take it this means you’ve run the new unit tests under `valgrind`? 🎉

pwithnall (commented):
Review of commit “language, diskimage: skip on unattended load error”, 👍.

pwithnall (commented):
Review of commit “finished: show unattended-specific error heading”, 👍.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -409,3 +409,109 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
 
   return GIS_UNATTENDED_COMPUTER_DOES_NOT_MATCH;
 }
+
+static GFile *
+invent_backup_name (GFile *source)
+{
+  g_autoptr(GFile) dir = g_file_get_parent (source);
```
pwithnall: Technically, `g_file_get_parent()` may return `NULL`, which isn’t handled below.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -409,3 +409,109 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
+                                gchar  **backup_path,
+                                GError **error)
+{
+  g_autoptr(GFile) target = invent_backup_name (source);
+  g_autoptr(GError) local_error = NULL;
+
+  if (g_file_move (source, target,
+                   /* No G_FILE_COPY_OVERWRITE, so target will not be
+                    * overwritten if it already exists. Since its name includes
+                    * a fresh UUID, this is extremely unlikely.
```
pwithnall: So if there *is* a `target` collision, this will error. Should it not retry instead?

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -409,3 +409,109 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
+
+  g_propagate_error (error, g_steal_pointer (&local_error));
+  return FALSE;
+}
+
+/**
+ * gis_unattended_config_write:
+ * @target_path: where to write new unattended config. If it already exists,
+ *  the existing file will be renamed to a unique name in the same directory.
+ * @backup_path: (out): set to the basename of backup of @target_path, if it
```
pwithnall: `(nullable)` and `(not optional)`

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -409,3 +409,109 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
+ *  already existed, or to %NULL if it did not.
+ *
+ * Either @vendor and @product must both be %NULL, or they must both be
+ * non-%NULL.
```
pwithnall: ⇒ `g_return_val_if_fail ((vendor == NULL) == (product == NULL), FALSE);`.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -409,3 +409,109 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
+    }
+  else
+    {
+      g_return_val_if_fail (product == NULL, FALSE);
+    }
+
+  if (!rename_existing_file_if_exists (target_file, backup_path, error))
+    return FALSE;
+
+  return g_key_file_save_to_file (key_file, target_path, error);
```
pwithnall: /me muses about the possibility of using `renameat2()` here to atomically exchange the temp file that the key file is saved to, and the (possibly already existing) `target_path`.

pwithnall: (This is not a serious suggestion.)

pwithnall (commented):
Review of commit “unattended: support writing config”, 🐈.

“GisUnattendedConfig is immutable once it has
been intialized, and I want to preserve that” — I salute you. 🖖

pwithnall (commented):
Review of commit “install: don’t unmount image partition”, 👍.

The code looks good to me, but I don’t know enough about the context to say whether this is a good idea in principle. Would it make sense to do a `syncfs()` call at this point instead, to further minimise the chance of an unclear unmount?

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -212,10 +215,73 @@ gis_finished_page_shown (GisPage *page)
...
+  g_autofree gchar *product = NULL;
+  g_autofree gchar *backup = NULL;
+  g_autoptr(GError) error = NULL;
+  GtkMessageType message_type = GTK_MESSAGE_INFO;
+  const gchar *message = NULL;
+  g_autofree gchar *secondary = NULL;
+  GtkWidget *dialog = NULL;
+
+  /* No-op if an error has occured. */
+  if (gis_store_get_error ())
```
pwithnall: Should we give the user some feedback in this case, like playing a ‘bzzt’ sound (in response to Ctrl+U being pressed)?

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -212,10 +215,73 @@ gis_finished_page_shown (GisPage *page)
...
+
+  if (gis_dmi_read_vendor_product (&vendor, &product, &error) &&
+      gis_unattended_config_write (unattended_ini, locale, image,
+                                   udisks_block_get_device (block),
+                                   vendor, product, &backup, &error))
+    {
+      message = _("Unattended installation configuration file created.");
+
+      if (backup != NULL)
+        secondary = g_strdup_printf (_("The previous file was renamed to %s."),
```
pwithnall: Nitpick: Add single quotation marks around the filename? `‘%s’`

pwithnall (commented):
Review of commit “finished: write unattended.ini when Ctrl+U is pressed”, 🗻.

gnome-image-installer/pages/confirm/Makefile.am
wjt: Almost certainly.

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: Great point

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: It should! But udisks (at least the version in our distro) does not support these macros. Workarounds would include using a variable of some weaker type, or defining the glue inline, but I think it's clearer to just `g_clear_object()` at the earliest possible point (which we do).

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: It does, sort of (by which I mean, it does not take a new reference, but nor does it release the reference).

gnome-image-installer/pages/confirm/gis-confirm-page.h
wjt: Yes, it's copy-pasta. On the other hand, I did make this page use `gtk_widget_class_set_template_from_resource` which no other page uses, so I could have updated this at the same time… oh well.

gnome-image-installer/pages/confirm/gis-confirm-page.ui
wjt: No, it's an in-joke placeholder that will always be replaced at runtime.

gnome-image-installer/pages/confirm/gis-confirm-page.c
pwithnall: Perhaps leave it as it is in this branch, but whack a `FIXME` on there and make sure there’s an upstream bug filed in udisks about adding `g_autoptr()` support (if not done already).

gnome-image-installer/pages/confirm/gis-confirm-page.c
pwithnall: But it does later unref what it’s passed, right?

gnome-image-installer/pages/confirm/gis-confirm-page.ui
pwithnall: The best.

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: Interesting – `UDisksDrive` comes from the output of `gdbus-codegen`, which can be configured to generate G_AUTO goop, but it does not appear to emit these definitions for the `UDisksDrive` interface, only for:

```c
G_DEFINE_AUTOPTR_CLEANUP_FUNC (UDisksDriveProxy, g_object_unref)
G_DEFINE_AUTOPTR_CLEANUP_FUNC (UDisksDriveSkeleton, g_object_unref)
```

So… maybe this is a GLib bug? :innocent:

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: Well… I think the answer is "yes". `GisPage` is a `GInitiallyUnowned`. It gets added to a `GtkStack`, and in response to `::destroyed` the GisAssistant (which GisDriver just hands the page over to) removes the pointer from the list. So I think the `GtkStack` consumes the initial floating reference and there are no other references.

These are all singletons so I'm not that worried anyway!

gnome-image-installer/pages/confirm/gis-confirm-page.c
pwithnall: More investigation required, if have a bit of time to take a look? I am currently down a GLib rabbit hole for ramcq.

gnome-image-installer/pages/confirm/gis-confirm-page.c
pwithnall: If you’re satisfied, I’ll leave it at that!

UNATTENDED.md
wjt: It could – I think it reads better too.

UNATTENDED.md
wjt: I changed both to "starting with" which is consistent with the discussion of `Computer` sections.

gnome-image-installer/util/gis-unattended-config.c
wjt: It's probably already (observably) idempotent as long as the underlying file doesn't change, though it does leak a few things. Ok.

wjt: https://developer.gnome.org/gio/stable/GInitable.html#g-initable-init has the following:

> Callers should not assume that a class which implements GInitable can be initialized multiple times, unless the class explicitly documents itself as supporting this. Generally, a class’ implementation of init() can assume (and assert) that it will only be called once. Previously, this documentation recommended all GInitable implementations should be idempotent; that recommendation was relaxed in GLib 2.54.

so I will not change this, particularly since the code using this class does not assume it can be initialized multiple times.

tests/Makefile.am
wjt: I'll leave this one because maybe one day, a new build system.

gnome-image-installer/util/gis-unattended-config.c
wjt: It is removed later in this patch series.

gnome-image-installer/gnome-image-installer.c
wjt: Good catch!

gnome-image-installer/util/gis-unattended-config.c
wjt: I reworked this function to make this case be covered by the regular success path.

gnome-image-installer/util/gis-dmi.c
wjt: Well, in theory it could have embedded NULs, but we already assume in many places that there are no NULs in the DMI data. The type is courtesy of `g_file_get_contents()` which fetches raw bytes but returns them as `gchar *`. I don't think introducing new types (and casts) will improve clarity, particularly given the function is documented to remove bytes (not characters) outside the ASCII printable range, but I'll try to reword the docs to make this clearer.

gnome-image-installer/util/gis-dmi.c
wjt: To me `126` seems more significant than `'~'` because I don't know the ASCII codepoint for tilde off-hand, but I do know that 127 is the highest codepoint defined in ASCII and is DEL.

gnome-image-installer/util/gis-unattended-config.c
pwithnall: …maybe add a comment briefly explaining this? :-)

gnome-image-installer/util/gis-dmi.c
pwithnall: 🤷‍♂️

gnome-image-installer/util/gis-unattended-config.c
wjt: Done, and required `image` to be non-empty if specified at the same time.

gnome-image-installer/util/gis-unattended-config.c
wjt: Fair.

gnome-image-installer/util/gis-unattended-config.c
wjt:
```
  gelf ~
% cat /tmp/foo.c
#define NULL ((void *) 0)

int main()
{
  char *arr[2] = { "abc", NULL };
  char **pp = &arr[0];
  const char **iter = pp;
  return iter[0] == NULL;
}
  gelf ~
% gcc -Wall -Wextra -o /tmp/foo /tmp/foo.c
/tmp/foo.c: In function ‘main’:
/tmp/foo.c:7:23: warning: initialization from incompatible pointer type [-Wincompatible-pointer-types]
   const char **iter = pp;
```

I don't think adding a cast would make this code more legible.

gnome-image-installer/util/gis-unattended-config.c
wjt: It should – at which point one may as well use `g_mkstemp()` (which has a retry loop), at which point maybe one may as well use `glnx_renameat2_exchange()`…

pwithnall: Well, yes, that sounds good.

gnome-image-installer/util/gis-unattended-config.c
wjt: This assertion is made below:

```c
  if (vendor != NULL)
    {
      g_return_val_if_fail (product != NULL, FALSE);
      // ...
    }
  else
    {
      g_return_val_if_fail (product == NULL, FALSE);
    }
```

This makes any assertion message more precise.

gnome-image-installer/util/gis-unattended-config.c
wjt: I took this suggestion.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: You're joking about the sound but I would like to draw your attention to https://github.com/endlessm/eos-shell/issues/4765

gnome-image-installer/util/gis-unattended-config.c
wjt: I added an assertion, as suggested in the documentation above.

gnome-image-installer/util/gis-unattended-config.c
wjt: Ok

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: TIL

gnome-image-installer/util/gis-unattended-config.c
pwithnall: :ohno:

gnome-image-installer/pages/finished/gis-finished-page.c
pwithnall: I’m actually not joking about the sound, although I probably should have better described it as ‘error bell’ or ‘terminal bell’.

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -416,66 +418,77 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
-                    * a fresh UUID, this is extremely unlikely.
-                    */
-                   G_FILE_COPY_NO_FALLBACK_FOR_MOVE,
-                   NULL, NULL, NULL, &local_error))
+  if (!g_file_query_exists (target_file, NULL))
     {
-      *backup_path = g_file_get_basename (target);
-      return TRUE;
+      *backup_basename = NULL;
+      return g_key_file_save_to_file (key_file, target_path, error);
```
pwithnall: It’s sad that there’s still a TOCTTOU race here, but no way around it since `g_key_file_save_to_file()` unconditionally overwrites things. :-(

gnome-image-installer/util/gis-unattended-config.c
```diff
@@ -416,66 +418,77 @@ gis_unattended_config_match_computer (GisUnattendedConfig *self,
...
-      backup_path = NULL;
-      return TRUE;
-    }
+  backup_path = invent_backup_name (target_path, error);
+  if (backup_path == NULL)
+    return FALSE;
 
-  g_propagate_error (error, g_steal_pointer (&local_error));
-  return FALSE;
+  if (!g_key_file_save_to_file (key_file, backup_path, error))
```
pwithnall: It’s also sad that we have to do I/O to invent a backup name, and then `g_key_file_save_to_file()` will do some very similar looking I/O to choose a temp filename to write to. Ideally we’d have the `renameat2()` call as part of `g_key_file_save_to_file()`, but that’s not a thing (and it’s definitely not worth rewriting this entire function to directly use POSIX I/O functions to achieve race-freedom).

pwithnall (commented):
Looks good to me. Two musings.

tests/test-unattended-config.c
```diff
@@ -249,15 +249,15 @@ test_missing_product (void)
 }
 
 static void
-test_full_blockdevice_path (void)
+test_full_block_devicepath (void)
```
pwithnall: s/`devicepath`/`device_path`/, or was that intentional?

pwithnall:
The fixup commits look good to me apart from one minor comment above (and two musings, but they are not actionable). The [build is failing though](https://ci.endlessm-sf.com/job/obs-eos-installer-pr/12/console):
```
ERROR: files left in build directory after distclean:
./gnome-image-installer/pages/confirm/confirm-resources.c
./gnome-image-installer/pages/confirm/confirm-resources.h
```

I think `$(BUILT_SOURCES)` is missing from `CLEANFILES` in `pages/confirm/Makefile.am` ⇒ one more fixup. After that, hopefully things are good to go.

pwithnall (changes requested):
r- as per comments above.

tests/test-unattended-config.c
wjt: It was not.

gnome-image-installer/pages/confirm/gis-confirm-page.c
pwithnall: Is that with `gdbus-codegen`’s `--c-generate-autocleanup` argument, or `--c-generate-autocleanup=all`? The default is `--c-generate-autocleanup=objects`.

See https://bugzilla.gnome.org/show_bug.cgi?id=763379 for the history behind this option.

gnome-image-installer/util/gis-unattended-config.c
wjt: To this and above, as discussed on Slack I don't think there is a way to atomically do "exchange if target exists, rename without overwriting if not", though one could run `renameat2(..., NO_REPLACE) || renameat2(..., EXCHANGE)` in a loop. One could also relatively easily open-code writing the keyfile to the tempfile's FD, and I got half-way through doing it, but I just don't think it's worth it. I agree that it is sad.

pwithnall (approved):
Looks good to me. Please squash and merge, pending results from Jenkins!

gnome-image-installer/pages/confirm/gis-confirm-page.c
wjt: Thanks for the pointer. udisks2 is using the default (ie, it does not pass `--c-generate-autocleanup` at all), and depends on too old a release of GLib (2.36) for this argument to be available. Joy! I'll open a PR there for discussion.

wjt: https://github.com/storaged-project/udisks/pull/492

PR #76, opened 2018-03-01
Improve and translate errors, run eos-diagnostics on error

wjt:
There's a bit of mostly-unrelated cleanup in here too.

https://phabricator.endlessm.com/T19950
https://phabricator.endlessm.com/T12822


pwithnall (commented):
Review of “diskimage: don't add unlocalized prefix to localized error”: :+1:

pwithnall (commented):
Review of “disktarget: don't manipulate list of pages”: :+1: :taco:

pwithnall (commented):
Review of “Update FSF's address”: :+1: :1st_place_medal:

pwithnall (commented):
Review of “Move error quark definitions to util”: :+1: :cheese:

pwithnall (commented):
Review of “confirm: remove unused GIS_CONFIRM_ERROR”: :+1: :hamburger:

pwithnall (commented):
Review of “configure: explicitly detect gpgconf”: :+1: :champagne:

pwithnall (commented):
Review of “Introduce and use better error codes”: :+1: :eagle:

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -206,16 +206,43 @@ gis_finished_page_shown (GisPage *page)
...
+        {
+          heading = _("Oops, something went wrong while finding suitable disks to reformat.");
+        }
+      else if (error->domain == GIS_INSTALL_ERROR)
+        {
+          heading = _("Oops, something went wrong while reformatting.");
+        }
+      else if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        {
+          heading = error->message;
```
pwithnall, line 281: I assume you’re OK with this (most likely) being `_("Operation was cancelled")`? There could potentially be issues here if someone does `g_set_error (error, G_IO_ERROR, G_IO_ERROR_CANCELLED, "some untranslated other message")`, but I think that’s unlikely.

pwithnall (commented):
Review of “finished: choose error headline based on GError.domain”: :+1:, albeit with one comment you might want to clarify

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -393,7 +393,7 @@ file_exists (
     return TRUE;
 
   g_set_error (error, GIS_IMAGE_ERROR, GIS_IMAGE_ERROR_NOT_FOUND,
-               "%s doesn't exist", path);
+               _("Image file ‘%s’ does not exist."), path);
```
pwithnall: Bonus points for use of “Unicode” “quotes”. :100: :1st_place_medal:

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -667,8 +667,8 @@ gis_scribe_write_thread_copy (GisScribe     *self,
     {
       g_autoptr(GError) local_error =
         g_error_new (GIS_IMAGE_ERROR, GIS_IMAGE_ERROR_WRONG_SIZE,
-                     "wrote %" G_GUINT64_FORMAT " bytes, "
-                     "expected to write %" G_GUINT64_FORMAT " bytes",
+                     _("Wrote %'" G_GUINT64_FORMAT " bytes, but "
+                       "expected to write %'" G_GUINT64_FORMAT " bytes."),
```
pwithnall: Is this what you were playing with `g_format_size_full()` for? Decided against using it in the end?

pwithnall (commented):
Review of commit “Mark more errors for translation; tweak style”: :+1:, albeit with one comment you might want to respond to. :tractor:

pwithnall (commented):
Review of “finished: hide Turn Off button on error in live session”: :+1:, although I can only review the code on this one, and not the design or motivation. Seems legit.

gnome-image-installer/gnome-image-installer.c
```diff
@@ -162,7 +162,7 @@ static gboolean
 check_for_live_boot (gchar **uuid)
 {
   const gchar *force = NULL;
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
```
pwithnall, line 165: Nitpick: I’ve been trying to be more consistent at using `local_error` as the name for a `GError*` local variable, and `error` as the name for a `GError**` parameter, since then you can add a `GError **error` parameter to an existing function without having to s/`error`/`local_error`/ its local variable first. However, what you have here is definitely correct, so this is more a comment for future reference.

pwithnall (commented):
Review for “Use g_debug/g_message/g_warning rather than g_print[err]”: :+1: :cocktail:

tests/test-scribe.c
```diff
@@ -254,46 +255,12 @@ fixture_set_up (Fixture *fixture,
...
-      g_warning ("Failed to delete %s: %s", path, error->message);
-      g_clear_error (&error);
-    }
-}
-
 static void
 fixture_tear_down (Fixture *fixture,
                    gconstpointer user_data)
 {
+  g_autoptr(GError) error = NULL;
```
pwithnall, line 262: Nitpick: As I think I mentioned on one of the earlier comments, the approach I’ve been taking recently has been to consistently call local `GError*`s `local_error`, to distinguish them from a possible `GError **error` parameter.

tests/test-unattended-config.c
```diff
@@ -38,47 +39,15 @@ fixture_set_up (Fixture *fixture,
...
-      g_clear_error (&error);
-    }
-}
-
 static void
 fixture_tear_down (Fixture *fixture,
                    gconstpointer user_data)
 {
-  rm_r (fixture->tmpdir);
+  g_autoptr(GError) error = NULL;
```
pwithnall, line 46: Nitpick: Same here.

pwithnall (commented):
Review of “tests: use glnx_shutil_rm_rf_at()”: :+1: :duck:

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -135,6 +140,54 @@ toggle_leds (GisPage *page)
...
 }
 
+static void
+write_diagnostics_cb (GObject      *source,
+                      GAsyncResult *result,
+                      gpointer      user_data)
+{
+  GisFinishedPage *self = GIS_FINISHED_PAGE (user_data);
+  GisFinishedPagePrivate *priv = gis_finished_page_get_instance_private (self);
+  g_autoptr(GError) error = NULL;
```
pwithnall: Nitpick: Same comment as in previous commits: I’d personally name this `local_error` rather than `error`.

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -135,6 +140,54 @@ toggle_leds (GisPage *page)
...
+      g_autofree gchar *basename_ = g_path_get_basename (output);
+      g_autofree gchar *link = NULL;
+      g_autofree gchar *markup = NULL;
+
+      if (priv->gedit != NULL)
+        {
+          g_assert (error == NULL);
+          g_autofree gchar *url = g_filename_to_uri (output, NULL, &error);
+
+          if (url == NULL)
```
pwithnall: If you returned a `GFile` from `gis_write_diagnostics_finish()`, you could skip all the error handling for converting the filename to a URI.

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -135,6 +140,54 @@ toggle_leds (GisPage *page)
...
+
+      if (link == NULL)
+        {
+          link = g_strdup (basename_);
+        }
+
+      markup = g_strdup_printf (_("A debug log has been saved as %s."), link);
+      gtk_label_set_markup (priv->diagnostics_label, markup);
+    }
+  else if (error != NULL)
```
pwithnall, line 177: Nitpick: Writing the `if`/`else if` this way implies to me that there could legitimately be a third condition `output == NULL && error == NULL` (otherwise why have an `else if` instead of an `else`?).

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -247,6 +301,18 @@ gis_finished_page_shown (GisPage *page)
...
 
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        {
+          /* Running eos-diagnostics can be slow enough that we can't do it
+           * synchronously on the UI thread.
+           */
+          GFile *image_dir = G_FILE (gis_store_get_object (GIS_STORE_IMAGE_DIR));
+          /* See implementation of gis_write_diagnostics_async for rationale. */
+          const gchar *home_dir = priv->gedit != NULL ? g_get_home_dir () : NULL;
+          gis_write_diagnostics_async (NULL, image_dir, home_dir,
```
pwithnall: What if this async operation takes longer than the remaining lifetime of `self`? Possible solutions:
 * Use `g_application_hold()` to keep the main loop running long enough to complete the operation, and keep a strong reference to `self`. You’d presumably allow the window to be unmapped while the operation finishes.
 * Keep a `GCancellable` in `self`, pass it to `gis_write_diagnostics_async()`, and cancel it when `self` is being `dispose`d.

Given that you want to write the diagnostic information at most costs, option 1 is probably better here.

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -346,6 +412,33 @@ write_unattended_config_cb (GisFinishedPage *self)
   gtk_widget_destroy (dialog);
 }
 
+static gboolean
+diagnostics_label_activate_link_cb (GtkLabel *label,
+                                    gchar    *uri,
```
pwithnall: `const`?

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -346,6 +412,33 @@ write_unattended_config_cb (GisFinishedPage *self)
...
+  GisFinishedPagePrivate *priv = gis_finished_page_get_instance_private (self);
+
+  /* We cannot rely on the default handler, because the FBE session is
+   * configured to associate a dummy handler (namely `/bin/true`) with all MIME
+   * types. So, explicitly launch gedit, provided it is installed.
+   */
+  if (priv->gedit != NULL)
+    {
+      g_autoptr(GList) uris = g_list_prepend (NULL, uri);
+      g_autoptr(GError) error = NULL;
```
pwithnall: Nitpick: `local_error` vs `error`. I’ll stop pointing this out now; by this point you either agree with me or have got sick of my nitpicks and wish I would shut up.

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -346,6 +412,33 @@ write_unattended_config_cb (GisFinishedPage *self)
...
+
+      if (g_app_info_launch_uris (G_APP_INFO (priv->gedit), uris, NULL,
+                                  &error))
+        return TRUE;
+
+      g_warning ("Failed to launch gedit for %s: %s", uri, error->message);
```
pwithnall: Should this `g_warning()` be outside the main `if` block in the function? i.e. Do you want to warn if gedit isn’t installed?

pwithnall: Although if gedit isn’t installed, there isn’t a hyperlink, so perhaps there should be a `g_assert_not_reached()` down there.

gnome-image-installer/pages/finished/gis-finished-page.c
pwithnall: Ah! There is a third condition. Maybe worth pointing that out in a comment here, since it’s fairly unusual.

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+
+  task = g_task_new (NULL, cancellable, callback, user_data);
+  g_task_set_task_data (task, d, (GDestroyNotify) write_diagnostics_data_free);
+  g_task_run_in_thread (task, write_diagnostics_thread_func);
+}
+
+/**
+ * gis_write_diagnostics_finish:
+ *
+ * Returns: the absolute path to the written diagnostics file; or %NULL with
```
pwithnall: `(transfer full) (nullable)`

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+
+  if (error != NULL)
+    g_task_return_error (task, g_steal_pointer (&error));
+  else
+    g_task_return_pointer (task, NULL, NULL);
+}
+
+/**
+ * gis_write_diagnostics_async:
+ * @eos_diagnostics_exe: (optional): executable to run to collect diagnostics
```
pwithnall, line 170: `(optional)` is not appropriate here (or on the other parameters for this function). `(optional)` means that it’s OK to pass `NULL` to an out-parameter to indicate that you don’ŧ want to receive a result via that out-parameter. `(nullable)` is what you want: it means it’s OK to pass `NULL` to an in-parameter to indicate that it should use a default value or be ignored (or whatever).

pwithnall: Aside: it’s valid to have `(optional) (nullable)` for an out-parameter which you can pass `NULL` to if you don’t want to receive a result from (`(optional)`); but if you do receive a result, that returned result could be `NULL` (`(nullable)`).

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+void
+gis_write_diagnostics_async (const gchar        *eos_diagnostics_exe,
+                             GFile              *image_dir,
+                             const gchar        *home_dir,
+                             GCancellable       *cancellable,
+                             GAsyncReadyCallback callback,
+                             gpointer            user_data)
+{
+  g_autoptr(GTask) task = NULL;
+  WriteDiagnosticsData *d =
```
pwithnall, line 189: Nitpick: I’d make this a `g_autoptr()`.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: Yes, I think that's fine. (Actually, at present the message can only be `_("Unattended installation was cancelled.")`; there are no GCancellables involved.)

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+   * location which is easily found on both Windows and Linux. The ideal place
+   * is the partition where the image is. This will fail if the image is on an
+   * ISO, since the ISO filesystem is read-only. It's also possible that the
+   * problem we hit is that we couldn't find the image partition!
+   */
+  if (d->image_dir != NULL)
+    {
+      g_autofree gchar *image_dirname = g_file_get_path (d->image_dir);
+      if (write_diagnostics_in_dir (task, image_dirname, output_basename,
+                                        diagnostics, &error))
```
pwithnall: Nitpick: Dodgy indentation here?

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+   * saving a log if the user can view it directly from the error page. (In
+   * particular, eosinstaller images do not include gedit or any other GUI text
+   * editor; so, in that case, if we couldn't save to the image partition we
+   * should just give up.)
+   */
+  if (d->home_dir != NULL)
+    {
+      g_clear_error (&error);
+      if (write_diagnostics_in_dir (task, d->home_dir, output_basename,
+                                        diagnostics, &error))
```
pwithnall: Nitpick: Dodgy indentation here?

gnome-image-installer/pages/install/gis-scribe.c
wjt: I didn't want to use `g_format_size_full()` because it only supports reporting the exact size in addition to (not instead of) the rounded size, ie “1.2 GB (1,234,567,890 bytes)”. I only want the second part, on the basis that the one time this actually happened on a correctly-signed image, the message would have been (with this change):

> wrote 101,241,834,086 bytes, expected to write 101,241,833,984 bytes

Whereas using `g_format_size_full (..., G_FORMAT_SIZE_LONG_FORMAT)` would give:

> wrote 101.2 GB (101,241,834,086 bytes), expected to write 101.2 GB (101,241,833,984 bytes)

The rounded size provides no useful information here!

(This condition can only happen if there is a bug in our image builder (as there was), because this check is performed after verifying the GPG signature.)

wjt: (But I did learn the `%'d` syntax from reading the `g_format_size_full()` source.)

gnome-image-installer/util/gis-write-diagnostics.c
```diff
@@ -0,0 +1,205 @@
...
+
+static WriteDiagnosticsData *
+write_diagnostics_data_new (const gchar *eos_diagnostics_exe,
+                            GFile       *image_dir,
+                            const gchar *home_dir)
+{
+  WriteDiagnosticsData *d = g_new (WriteDiagnosticsData, 1);
+  d->eos_diagnostics_exe = g_strdup (eos_diagnostics_exe ?: "eos-diagnostics");
+  d->image_dir = image_dir != NULL ? g_object_ref (image_dir) : NULL;
+  d->home_dir = g_strdup (home_dir ?: NULL);
```
pwithnall: The ternary seems a bit pointless here, since `home_dir` is either set or `NULL`.

pwithnall: Unless it’s there to make it more obvious that `home_dir` _could_ be `NULL`, in which case: fine, carry on.

tests/test-write-diagnostics.c
```diff
@@ -0,0 +1,287 @@
...
+
+  gchar *output;
+  GError *error;
+} Fixture;
+
+static void
+fixture_set_up (Fixture *fixture,
+                gconstpointer user_data)
+{
+  GError *error = NULL;
```
pwithnall, line 42: Nitpick: `g_autoptr()` for consistency and documentation of ownership, even if we will always abort if the error is ever set.

tests/test-write-diagnostics.c
```diff
@@ -0,0 +1,287 @@
...
+}
+
+static gboolean
+mkdir_ (const char *path,
+        GError    **error)
+{
+  mode_t mode = 0755;
+
+  if (g_mkdir (path, mode) != 0)
+    return glnx_throw_errno_prefix (error, "mkdir(%s, %o)", path, mode);
```
pwithnall: Why not use `glnx_ensure_dir (-1, path, 0755, error)`? Does this need to work on WIndows?

pwithnall (commented):
Review of “finished: run eos-diagnostics on error”: A few comments which could do with a reply, but no major changes. Basically :+1:.

gnome-image-installer/pages/finished/gis-finished-page.c
```diff
@@ -185,6 +185,7 @@ write_diagnostics_cb (GObject      *source,
     {
       g_prefix_error (&error, _("Failed to save debug log: "));
       gtk_label_set_text (priv->diagnostics_label, error->message);
+      g_message ("%s: %s", G_STRFUNC, error->message);
```
pwithnall: Why not `g_warning()` instead? (Similarly below.)

pwithnall (commented):
Review of “finished: ensure errors end up in the journal”: One question needs answering. :-1:

pwithnall (commented):
Review of “glnx: update snapshot”: Trivially yes. :+1:  :confetti_ball: :fireworks:

gnome-image-installer/pages/install/gis-scribe.c
pwithnall: You may potentially want to put that in a comment by the code, but the need for it is a bit borderline.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: ok.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: Fair. Yes, it's unusual, but it worked out neatest to allow both arguments to be NULL.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: ok.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: guess so

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: I don't want to warn if gedit isn't installed. This is the expected case on `eosinstaller` images.

gnome-image-installer/util/gis-write-diagnostics.c
wjt: Why? It's task data, with its lifetime tied to the GTask.

gnome-image-installer/util/gis-write-diagnostics.c
wjt: extremely.

gnome-image-installer/util/gis-write-diagnostics.c
wjt: This was leftover from another incarnation of this function I think.

tests/test-write-diagnostics.c
wjt: I did not know that function existed.

gnome-image-installer/pages/finished/gis-finished-page.c
wjt: I remembered why this wasn't const: `g_list_prepend` takes a non-const argument, so now I have to cast the const away again! But ok.

gnome-image-installer/util/gis-write-diagnostics.c
pwithnall: Not until you call `g_task_set_task_data()`.

tests/test-write-diagnostics.c
pwithnall: libglnx has wrappers for most of the commonly-used POSIX functions now, adding `GError` support and `EINTR` hackery.

pwithnall:
The three fixup commits look good to me. I haven’t looked back to check they address all my womblings; I assume they do. If so, I’m fine with this being squashed and merged!

pwithnall (approved):

PR #77, opened 2018-06-07
Add extra checks and logging to scribe; fix many warnings

wjt:
I saw a transient GPG verification failure when reformatting from a nightly ISO in a virtual machine. There was no useful information in the journal (beyond GPG's "BAD signature" output); re-running the process (without a reboot or anything) worked fine; and I manually GPG-verified the ISO, the squashfs within it, and the uncompressed image within that. Mysterious!

There isn't really much to go on. Maybe this was a transient read error from the USB disk on which the ISO is stored. Maybe a cosmic ray flipped a bit. But in case it happens again, here is some extra debug output and an extra coherence check in the scribe.

I also included many warning fixes, found by editing this project in GNOME Builder building against a non-Endless system.

### DIFF
--- a/gnome-image-installer/pages/install/gis-scribe.c
+++ b/gnome-image-installer/pages/install/gis-scribe.c
@@ -1072,7 +1072,7 @@ gis_scribe_begin_verify (GisScribe *self,
     g_pollable_input_stream_create_source (G_POLLABLE_INPUT_STREAM (gpg_stdout),
                                            cancellable);
   g_task_attach_source (task, task_data->stdout_source,
-                        (GSourceFunc) gis_scribe_gpg_progress);
+                        (GSourceFunc) (void (*)(void)) gis_scribe_gpg_progress);
 
   g_subprocess_wait_check_async (task_data->subprocess, cancellable,
                                  gis_scribe_gpg_wait_check_cb,
--- a/gnome-image-installer/pages/install/gis-scribe.c
+++ b/gnome-image-installer/pages/install/gis-scribe.c
@@ -1034,6 +1034,7 @@ gis_scribe_begin_verify (GisScribe *self,
       "--input-size-hint", size_str,
       "--verify", signature_path, "-", NULL
   };
+  g_autofree gchar *args_flat = g_strjoinv (" ", (gchar **) args);
   g_autoptr(GSubprocessLauncher) launcher = NULL;
   GInputStream *gpg_stdout = NULL;
   GOutputStream *gpg_stdin = NULL;
@@ -1052,6 +1053,7 @@ gis_scribe_begin_verify (GisScribe *self,
       return NULL;
     }
 
+  g_message ("Spawning %s", args_flat);
   launcher = g_subprocess_launcher_new (G_SUBPROCESS_FLAGS_STDIN_PIPE |
                                         G_SUBPROCESS_FLAGS_STDOUT_PIPE);
   task_data->subprocess = g_subprocess_launcher_spawnv (launcher, args, &error);
--- a/gnome-image-installer/pages/install/gis-scribe.c
+++ b/gnome-image-installer/pages/install/gis-scribe.c
@@ -509,6 +509,38 @@ gis_scribe_new (GFile       *image,
 }
 
 static void
+task_return_error (GisScribe *self,
+                   GTask     *task,
+                   GError    *error)
+{
+  g_mutex_lock (&self->mutex);
+  if (self->error == NULL)
+    self->error = g_error_copy (error);
+  g_mutex_unlock (&self->mutex);
+
+  g_task_return_error (task, g_steal_pointer (&error));
+}
+
+G_GNUC_PRINTF (5, 6)
+static void
+task_return_new_error (GisScribe   *self,
+                       GTask       *task,
+                       GQuark       domain,
+                       gint         code,
+                       const gchar *format,
+                       ...)
+{
+  va_list args;
+  GError *error;
+
+  va_start (args, format);
+  error = g_error_new_valist (domain, code, format, args);
+  va_end (args);
+
+  task_return_error (self, task, g_steal_pointer (&error));
+}
+
+static void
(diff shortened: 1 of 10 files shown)


gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1072,7 +1072,7 @@ gis_scribe_begin_verify (GisScribe *self,
     g_pollable_input_stream_create_source (G_POLLABLE_INPUT_STREAM (gpg_stdout),
                                            cancellable);
   g_task_attach_source (task, task_data->stdout_source,
-                        (GSourceFunc) gis_scribe_gpg_progress);
+                        (GSourceFunc) (void (*)(void)) gis_scribe_gpg_progress);
```
pwithnall, line 1152: I think upstream would probably be interested in a `G_SOURCE_FUNC()` casting macro for this. But to avoid dependency faff, you should go forth with `(void (*)(void))` here for now.

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1110,6 +1110,8 @@ gis_scribe_tee_data_free (GisScribeTeeData *data)
   g_slice_free (GisScribeTeeData, data);
 }
 
+G_STATIC_ASSERT (G_MAXSSIZE <= G_MAXUINT64);
```
pwithnall: What does this relate to? It’s not immediately obvious (from the diff view in GitHub) where you’re sticking a `gsize` value in a `guint64`.

⇒ Maybe add a comment

wjt: It's in the function below but I think this check is (a) unnecessary (our images are certainly < 2 ** 62 - 1) or (b) incomplete (the gsize is repeatedly added to the guint64 so if I were really over-egging it I should add an overflow check). OK.

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1147,9 +1149,18 @@ gis_scribe_tee_thread (GTask            *task,
           g_prefix_error (&error, "error writing image to self: ");
           break;
         }
+
+      bytes_teed += r;
```
wjt, line 1230: I realised that 2 out of the 3 error paths above are not reached in the test suite, so I'll take a look at fixing that…

wjt: The read and write-to-GPG paths are now tested (and worked all along).

tests/test-scribe.c
```diff
@@ -787,6 +790,33 @@ main (int argc, char *argv[])
...
+  g_test_add ("/scribe/gpg-exits-prematurely/success", Fixture, &gpg_exits_prematurely_success,
+              fixture_set_up,
+              test_error,
+              fixture_tear_down);
+
+  TestData gpg_exits_prematurely_failure = {
+      .image_path = image_path,
+      .signature_path = image_sig_path,
+      .gpg_path = "false",
+      .error_domain = G_IO_ERROR,
```
wjt: Oh right! This is racy! If the tee task notices its pipe-to-GPG is closed before the subprocess task notices "GPG" has exited, we'll get BROKEN_PIPE. But if (as happened) on Jenkins the race goes the other way, we'll get GIS_IMAGE_ERROR_VERIFICATION_FAILED. I think I don't care which error we get, just that *some* error is reported.

gnome-image-installer/pages/install/gis-scribe.c
pwithnall: *hopeful smile*

gnome-image-installer/pages/install/gis-scribe.c
pwithnall: Yeah, if you’re repeatedly adding the `gsize` to the `guint64` then it needs a runtime overflow check, if you care about that kind of thing.

gnome-image-installer/pages/install/gis-scribe.c
wjt: MVP at https://gitlab.gnome.org/GNOME/glib/merge_requests/82

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -228,6 +230,11 @@ gis_scribe_set_property (GObject      *object,
       self->convert_to_mbr = g_value_get_boolean (value);
       break;
 
+    case PROP_GPG_PATH:
+      g_free (self->gpg_path);
+      self->gpg_path = g_value_dup_string (value);
```
pwithnall, line 235: Nitpick: Maybe add some assertions that the path is non-empty and absolute? I normally add such assertions in the `constructed` vfunc, since it’s guaranteed to be called after all construct-only properties are set, regardless of whether a value is actually provided for the property (or the default value is used).

tests/test-scribe.c
```diff
@@ -372,9 +375,14 @@ test_error (Fixture       *fixture,
...
-                  fixture->data->error_domain,
-                  fixture->data->error_code);
+
+  /* In some cases we don't care what the error is, just that there is an
+   * error.
+   */
+  if (fixture->data->error_domain != 0)
+    g_assert_error (error,
+                    fixture->data->error_domain,
+                    fixture->data->error_code);
```
pwithnall, line 385: Maybe add
```
else
  g_assert_nonnull (error);
```
?

pwithnall (commented):
A couple of small comments.

gnome-image-installer/pages/install/gis-scribe.c
wjt: There's a non-NULL assertion in `constructed`. I'll add a check for absolution, and make it hold true.

tests/test-scribe.c
wjt: Oh yes! I meant to do this. Thanks.

pwithnall (approved):
LGTM!

PR #79, opened 2019-05-31
scribe: work around xgettext bug

wjt:
The previous string was extracted as:

    #: ../gnome-image-installer/pages/install/gis-scribe.c:754
    msgid "Wrote %'"
    msgstr ""

Such is life.


treitter (approved):
Looks good, assuming this doesn't trigger the previously-seen xgettext bug

wjt:
```
#: ../gnome-image-installer/pages/install/gis-scribe.c:768
#, c-format
msgid "Wrote %s bytes, but expected to write %s bytes."
msgstr ""
```

PR #80, opened 2019-07-01
diskimage: remove fnde special-cases

wjt:
This never made it out of a pilot deployment.


pwithnall:
Nice. For context, the most recent ping from an installed system was from May 2018.

PR #82, opened 2019-09-24
diskimage: log and skip mount failures

dsd:
With the extra data partition newly added (but not yet formatted) in our
ISO-USB boot path, we now face a new complication finding the partition
that contains the image.

Due to some awkward blkid and udev behaviour, the unformatted extra
data partition will inherit the label and UUID of the main iso9660
partition. If the installer finds this as the image partition, it will
fail to mount, and the installer will abort at that point.

Avoid this situation by logging mount failures but then continuing
on to try any other appropriate devices.

To avoid a fairly significant rework which we want to avoid at this time,
the mounting had to be made synchronous in order to easily detect and
skip the mount error condition.

https://phabricator.endlessm.com/T27158

### DIFF
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -548,23 +548,6 @@ gis_diskimage_page_populate_model (GisPage     *page,
 }
 
 static void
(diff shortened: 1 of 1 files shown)


wjt (approved):
LGTM

PR #83, opened 2019-10-21
PolicyKit rules changes

ptomato:
These should avoid inadvertently blocking the gnome-initial-setup rules.

https://phabricator.endlessm.com/T28267


eos-installer-data/90-eos-installer.rules
```diff
@@ -19,22 +19,29 @@ try {
 
 
 polkit.addRule(function(action, subject) {
-    if (subject.user !== 'gnome-initial-setup' && !(subject.user === 'live' && isLiveSystem))
+    // This rule should be identical to the one in 20-gnome-initial-setup.rules,
+    // to allow running the installer as the live user instead of the
+    // gnome-initial-setup user.
```
wjt: I think this entire function can be dropped – the `eos-installer` application doesn't need any of these permissions. It is a bug that they are present.

wjt (changes requested):

eos-installer-data/90-eos-installer.rules
ptomato: Are you sure? I was thinking that it was set up this way in commit 0fafce9ed because when you are on a live boot and then run the eos-installer you are the live user, and not the gnome-initial-setup user, and so the live user has to have the same accesses as gnome-initial-setup. Or does that not apply because the restriction of "can't show an authorization dialog" (which makes this necessary for gnome-initial-setup) doesn't apply to the live user?

wjt:
```
Making all in eos-installer-data
make[2]: Entering directory '/build/src/eos-installer-data'
sed -e 's|@SBINDIR[@]|/usr/local/sbin|g' -e 's|@SYSCONFDIR[@]|/usr/local/etc|g' -e 's|@LIBDIR[@]|/usr/local/lib|g' -e 's|@LIBEXECDIR[@]|/usr/local/libexec|g' -e 's|@LOCALSTATEDIR[@]|/usr/local/var|g' -e 's|@DATADIR[@]|/usr/local/share|g' com.endlessm.Installer.desktop.in.in >com.endlessm.Installer.desktop.in
make[2]: *** No rule to make target '20-eos-installer.rules', needed by 'all-am'.  Stop.
```

eos-installer-data/90-eos-installer.rules
wjt: I believe that the only Polkit actions that `eos-installer` needs to take (ignoring which user it is run as) are:

1. enumerate devices with UDisks, possibly mounting some (polkit action), and opening the target device for writing (polkit action)
2. running `/usr/sbin/eos-repartition-mbr` as the root user with `pkexec` (polkit action)

It doesn't need any of the other permissions currently listed in this file. They are here because it is a fork of `gnome-initial-setup` (the application), which does need them all. For a long time it did use the language page, but I deleted that (and all other `gnome-initial-setup` pages) in favour of letting `gnome-initial-setup` set the language.

ptomato: > I deleted that (and all other gnome-initial-setup pages)

That was the bit of information that I was missing, thanks!

wjt (approved):

PR #86, opened 2020-04-08
Fix path to xattr.h header

valentindavid:
`xattr.h` has been moved into glibc.


wjt:
ok to test

wjt:
@valentindavid thanks!

PR #87, opened 2020-04-08
Allow images without separate /efi & /boot partitions

valentindavid:
This is for GNOME OS where we use systemd boot where we do not require
the use of a separate partition.

https://phabricator.endlessm.com/T30030

PR #90, opened 2021-03-23
Rewrite README

wjt:
Apparently I never updated this to talk about ISO images. There were
many other missing details that I added while I was here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,45 +1,140 @@
 Endless OS Reformatter
 ======================
 
+![](./eos-installer-data/icons/hicolor/64x64/apps/com.endlessm.Installer.png)
+
 This application is run from a live USB to write an Endless OS disk image to a
 fixed disk. It is branded as “Reformat with Endless OS” or “Reformat” to avoid
-confusion with the Endless Installer for Windows.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,45 +1,130 @@
...
+--------------------------------
 
+We also publish an `eosinstaller` ostree & OS image. This is a cut-down build
+of the OS which excludes almost all the normal applications: it can only be
+used to run eos-installer (this application) from the `gnome-initial-setup`
+greeter session.
 
-Modes
------
+The expected partition layout in this case is: EFI and BIOS boot partitions, an
```
dbnicholson: Is this still accurate after #87? As far as I can tell, it's just an EFI partition + a root partition of some type.

dbnicholson (commented):
Thanks, this is very helpful. I just had a question about a particular.

README.md
wjt: Actually this is still technically correct because it refers to the partition layout of the `eosinstaller` image, not the OS image it is installing, but I'll reword to make this clearer.

dbnicholson (approved):

PR #91, opened 2021-03-30
Verify with checksum when signature not available

dbnicholson:
This needs a little polish but I wanted to get some feedback. If there's no `.asc` file next to the image, try looking for a `.sha256` file and do a checksum verification.

https://phabricator.endlessm.com/T31304


gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1150,6 +1186,128 @@ gis_scribe_begin_verify (GisScribe *self,
...
+          _("The checksum file ‘%s’ does not exist."),
+          checksum_path);
+      return NULL;
+    }
+
+  /* Read in the checksum file to get the expected checksum. */
+  if (!g_file_load_contents (self->checksum, cancellable, &checksum_contents,
+                             &checksum_len, NULL, &error))
+    {
+      task_return_error (self, task, g_steal_pointer (&error));
```
dbnicholson, line 1268: I think this is an existing bug, but currently this will deadlock the installer using `task_return_error` because the `gis_scribe_begin_*` functions used for task setup are called with the mutex locked. I briefly changed them back to `g_task_return_error`, but then the error message that shows up is wrong because it doesn't set `GisScribe::error`. I'm not sure what the best approach is.

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
```diff
@@ -117,6 +119,12 @@ gis_diskimage_page_selection_changed(GtkWidget *combo, GisPage *page)
   gis_store_set_image_signature (signature);
   g_free (signature);
 
+  if (checksum == NULL)
+    checksum = g_strjoin (NULL, image, ".sha256", NULL);
+
+  gis_store_set_image_checksum (checksum);
+  g_free (checksum);
```
wjt: These days I would declare this variable as:

```
g_autofree char *checksum = NULL;
```

and then not need the manual `g_free`.

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1507,11 +1667,32 @@ gis_scribe_write_async (GisScribe          *self,
...
+    }
+  else
+    {
+      g_autofree gchar *signature_path = g_file_get_path (self->signature);
+      g_autofree gchar *checksum_path = g_file_get_path (self->checksum);
+      g_task_return_new_error (
+          task, GIS_IMAGE_ERROR, GIS_IMAGE_ERROR_VERIFICATION_FAILED,
+          _("Neither the signature file ‘%s’ nor the checksum file ‘%s’ exist."),
+          signature_path, checksum_path);
+      goto out;
```
wjt: Won't this skip the block just below that closes the other pipes?

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1150,6 +1186,128 @@ gis_scribe_begin_verify (GisScribe *self,
...
+    }
+
+  if (checksum_len < CHECKSUM_STRLEN)
+    {
+      task_return_new_error (
+          self, task, GIS_IMAGE_ERROR, GIS_IMAGE_ERROR_VERIFICATION_FAILED,
+          _("The checksum file ‘%s’ does not contain %d characters."),
+          checksum_path, CHECKSUM_STRLEN);
+      return NULL;
+    }
```
wjt: Seems redundant with the block below?

gnome-image-installer/pages/install/gis-scribe.c
```diff
@@ -1150,6 +1186,128 @@ gis_scribe_begin_verify (GisScribe *self,
...
+
+  for (cur = checksum_contents; cur && *cur; cur++)
+    {
+      if (!g_ascii_isxdigit (*cur)) {
+        task_return_new_error (
+            self, task, GIS_IMAGE_ERROR, GIS_IMAGE_ERROR_VERIFICATION_FAILED,
+            _("The checksum file ‘%s’ contains invalid character ‘%c’."),
+            checksum_path, *cur);
```
wjt: `*cur` might be part of a multi-byte UTF-8 sequence. I wonder what would be substituted into the string in that case?

wjt (commented):
Had a couple of superficial comments but structurally this seems fine. Next stop, https://github.com/endlessm/eos-image-builder/pull/2.

gnome-image-installer/pages/install/gis-scribe.c
wjt: Ah, I'd forgotten this particular wart. I was never too happy with how the subtasks' errors propagate upwards. I'm not sure if there is more advanced technology for multiplexed GTask these days.

Perhaps the different fields could be guarded by different mutexes? Looking at the code, I think `gis_scribe_write_thread_await_gpg()` could be made to only access `self->outstanding_tasks`…

gnome-image-installer/pages/diskimage/gis-diskimage-page.c
dbnicholson: Yes, that would be better. `signature` above is doing the same thing.

gnome-image-installer/pages/install/gis-scribe.c
dbnicholson: Oh, good point. I don't think I fully understand the error handling here.

gnome-image-installer/pages/install/gis-scribe.c
dbnicholson: Right. I had something different originally but this is unnecessary.

gnome-image-installer/pages/install/gis-scribe.c
dbnicholson: You could also change the `outstanding_tasks` usage to atomic integer access, but I'm not sure that's the right model since the top level seems like it wants to keep the lock until all the tasks are setup.

wjt: Oh, that could work. I think I wanted to have the property that it would always be non-zero at the end of this function, so that one of the earlier subtasks failing couldn't cause the task we're setting up to be completed while we're still going.

I don't actually know if that's possible. And it could be faked by adding an extra field to the bitfield for "SETUP".

gnome-image-installer/pages/install/gis-scribe.c
dbnicholson: Garbage? I didn't try to handle this, but let me know what you think.

dbnicholson:
OK, I think this is ready now. The first commit is a bug fix (with tests in the 2nd commit) which I can split out to a separate PR if you prefer.

gnome-image-installer/pages/install/gis-scribe.c
dbnicholson: I ended up just locking and unlocking the mutex around each access to `outstanding_tasks` in `gis_scribe_write_async`. Initially I thought it was important that the mutex was held until all the subtasks were started, but I don't think that's necessary. I added tests for this (and manually tested by running the installer with no signature present) and it did the right thing.

wjt: OK, makes sense!

gnome-image-installer/pages/install/gis-scribe.c
wjt: It's probably fine as-is.

wjt (approved):
Looks great, thanks!

PR #92, opened 2021-04-14
Allow loop device for installation target

dbnicholson:
The requirement to have an actual non-removable disk on hand for testing is pretty onerous. This allows using a loop device for the installation target and ensures that it's not the same if the source is also a loop device.

I feel like you could probably do some automated testing with this presuming that the test was run from a fairly privileged process.

### DIFF
--- a/gnome-image-installer/pages/disktarget/gis-disktarget-page.c
+++ b/gnome-image-installer/pages/disktarget/gis-disktarget-page.c
@@ -64,6 +64,7 @@ check_can_continue(GisDiskTargetPage *page)
   GisDiskTargetPagePrivate *priv = gis_disktarget_page_get_instance_private (page);
   UDisksBlock *block = UDISKS_BLOCK (gis_store_get_object (GIS_STORE_BLOCK_DEVICE));
   UDisksDrive *drive = NULL;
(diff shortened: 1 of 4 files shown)


gnome-image-installer/pages/disktarget/gis-disktarget-page.c
```diff
@@ -287,51 +287,79 @@ gis_disktarget_page_populate_model(GisPage *page, UDisksClient *client)
...
+          skip_if (block == NULL, "no corresponding block object");
+
+          if (udisks_block_get_size (block) >= gis_store_get_required_size ())
+            {
+              priv->has_valid_disks = TRUE;
+            }
+
+          targetname = g_strdup_printf("%s %s",
+                                       udisks_block_get_device (block),
+                                       udisks_loop_get_backing_file (loop));
```
wjt: You can use `backing_file` here

wjt (approved):
LGTM, besides one superficial change.

All this logic (including the existing logic) feels a bit messy, but this seems to be the fault of the underlying APIs (UDisks and below).

dbnicholson:
I made that change and added some documentation on using loop devices for testing.

PR #93, opened 2021-04-20
Skip unwanted loop targets

dbnicholson:
This strips out more loop devices as potential installation targets. Without this, the loop devices holding the image and squashfs in the ISO are shown to the user.

https://phabricator.endlessm.com/T31304

### DIFF
--- a/gnome-image-installer/pages/disktarget/gis-disktarget-page.c
+++ b/gnome-image-installer/pages/disktarget/gis-disktarget-page.c
@@ -359,6 +359,8 @@ gis_disktarget_page_populate_model(GisPage *page, UDisksClient *client)
                                            G_FORMAT_SIZE_DEFAULT);
         }
 
+      skip_if (udisks_block_get_read_only (block), "block device is read-only");
+
       block_device = udisks_block_get_device (block);
       skip_if (config != NULL &&
                !gis_unattended_config_matches_device (config, block_device),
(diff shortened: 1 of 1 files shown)


wjt (approved):

gnome-image-installer/pages/disktarget/gis-disktarget-page.c
```diff
@@ -340,6 +340,8 @@ gis_disktarget_page_populate_model(GisPage *page, UDisksClient *client)
           const gchar *backing_file = udisks_loop_get_backing_file (loop);
           skip_if (backing_file == NULL || *backing_file == '\0',
                    "no backing file");
+          skip_if (!g_file_test (backing_file, G_FILE_TEST_EXISTS),
+                   "backing file %s does not exist", backing_file);
```
wjt, line 344: Having merged this I guess there is actually a problem here. If the file is not visible to the user running eos-installer, then it will appear to not exist.

The instructions in the README would not lead to this situation since they encourage you to create the files as your own user. So it's probably OK.

dbnicholson: Good point. I had left this off initially but got a little worried that the loop wouldn't be marked read-only at some point. For instance, if you just run `mount something.iso /somewhere`, mount will setup the loop for you, but it doesn't make it read-only (you can check with `losetup -la`). I don't know how the loops are setup in the running ISO, so I was worried that would be changed at some point.

I agree that this is probably OK, though, since the loop handling is just for testing and hopefully the message would be clear enough for someone doing that testing.

PR #95, opened 2021-12-09
diskimage: find names for uncompressed image files

alyssais:
Uncompressed image files with names ending in .img are already
accepted by add_image(), but the regex for finding the display name
would fail because it didn't understand files without a second
extension after .img.

Uncompressed image files are very useful for my use case: by using a
loopback mount in GRUB (which doesn't work if the image file is
compressed), I can provide the option of booting the image directly
from GRUB as a live image.  The OS I'm working with is special
purpose, and it wouldn't make sense to run eos-installer on it
directly, so eos-installer runs on a standalone standard Linux image
created just for the installer, instead of being part of the live
image like it normally would be.


dbnicholson:
ok to test

dbnicholson (approved):
Makes sense!

PR #96, opened 2021-12-09
Add more log messages for invalid disk images

alyssais:
These messages would all have been extremely helpful to me when I was
trying to set up eos-installer and wondering why it didn't like my
images!


dbnicholson:
ok to test

dbnicholson (approved):
Yeah, that's helpful.

PR #97, opened 2021-12-09
README: document uncompressed images and sha256

alyssais:


wjt:
Good catch. Thanks – I've added a few more references to SHA256.

PR #98, opened 2022-01-06
Support RISC-V disk images

wjt:
This is for the benefit of GNOME OS, which uses eos-installer and supports RISC-V, and is based on a patch in GNOME OS.


wjt:
@valentindavid How does this look? I was browsing https://gitlab.gnome.org/GNOME/gnome-build-meta/-/tree/master/files/eos-installer for reasons I can't remember and this one seemed easy enough to sort out.

valentindavid:
This looks correct, yes. Thank you. Seems I forgot to submit this one.

wjt:
No worries!

If you are ever filled with enthusiasm to submit the branding-related patches (configure flag? conditional based on `/usr/lib/os-release`?) that'd be great. I imagine based on @alyssais's recent contributions that generic/parameterised branding would also be useful for her work.

PR #99, opened 2022-05-03
UNATTENDED: Ctrl-click link instead of shift-click

cassidyjames:
I at least know Ctrl+Click works, and it's what we have documented at https://support.endlessos.org/en/deployment/reformatter-usb. I haven't tested if Shift+Click also works, but it seems we should at least be consistent.


UNATTENDED.md
```diff
@@ -2,9 +2,9 @@
...
 
-Once you have a reformatter USB, the easiest way to configure it for unattended mode is to run it in the normal, attended mode on your target system. Once the installation is complete, press `Ctrl+U` when you reach the “Reformatting Finished” page. This will write an `unattended.ini` configuration file matching this system and the options you chose; you can now use this USB stick to reformat other computers (with the same manufacturer and model) without any user interaction.
+Once you have a reformatter USB, the easiest way to configure it for unattended mode is to run it in the normal, attended mode on your target system. Once the installation is complete, press <kbd>Ctrl</kbd><kbd>U</kbd> when you reach the “Reformatting Finished” page. This will write an `unattended.ini` configuration file matching this system and the options you chose; you can now use this USB stick to reformat other computers (with the same manufacturer and model) without any user interaction.
```
wjt:
```suggestion
Once you have a reformatter USB, the easiest way to configure it for unattended mode is to run it in the normal, attended mode on your target system. Once the installation is complete, press <kbd>Ctrl</kbd>+<kbd>U</kbd> when you reach the “Reformatting Finished” page. This will write an `unattended.ini` configuration file matching this system and the options you chose; you can now use this USB stick to reformat other computers (with the same manufacturer and model) without any user interaction.
```

wjt (approved):
Huh. Yep, only Ctrl + click works:

https://github.com/endlessm/rufus/blob/692fd09a1800cb77c2094487c7a15842920b1786/endless/src/endless/EndlessUsbToolDlg.cpp#L2141-L2142

I believe that using + between keys when describing key combinations is pretty standard. Otherwise LGTM

PR #100, opened 2022-07-18
polkit rules: use non-sugary closure syntax

alatiera:
Polkit added a ducktape backend in 121 release,
and it looks like it doesn't support the => closures

### DIFF
--- a/eos-installer-data/90-eos-installer.rules
+++ b/eos-installer-data/90-eos-installer.rules
@@ -34,7 +34,7 @@ polkit.addRule(function(action, subject) {
     return undefined;
 });
 
-polkit.addRule((action, subject) => {
+polkit.addRule(function(action, subject) {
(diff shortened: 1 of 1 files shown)


alatiera:
PS: I haven't yet verified that ducktape does indeed not implement `=>` even though it seems like the likely explanation. I will try to find a reference for it once done debugging

wjt (approved):
This change seems fine anyway, it's consistent with the other rule in the same file.

alatiera:
For future ref here is the duktape issue https://github.com/svaarala/duktape/issues/275

nanonyme:
@wjt FWIW seems matter boils down to that mozjs has sneakily been increasing allowed JavaScript way past what is allowed by polkit spec. The problem is not that duktape doesn't support this syntax, it is not allowed at all.

PR #101, opened 2022-07-19
polkit rules: tweak to be ecma 5 compatible

alatiera:
Or how I lost a weekend debugging..

Polkit rules must be ecma 5 compatible [1]

* Use var instead of const for the array
* Use indexOf instead of includes

[1] https://gitlab.freedesktop.org/polkit/polkit/-/blob/master/docs/man/polkit.xml#L504

### DIFF
--- a/eos-installer-data/90-eos-installer.rules
+++ b/eos-installer-data/90-eos-installer.rules
@@ -43,13 +43,13 @@ polkit.addRule(function(action, subject) {
         !(subject.user === 'live' && isLiveSystem))
         return undefined;
 
-    const allowedUdisksActions = [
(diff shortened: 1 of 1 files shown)


wjt:
hey jenkins! this is ok to test

wjt:
LGTM – I don't suppose you know of a decent way to lint these rules?

wjt:
Is this a draft because you want to do some further testing?

alatiera:
nope, there doesn't seem a way to validate them before hand, I am running a test build now.

alatiera:
It works :tada:

wjt (approved):

PR #103, opened 2022-10-10
Modernize packaging

wjt:
Sort build dependencies; use debhelper rather than CDBS; and update our contact details.

https://phabricator.endlessm.com/T33912


dbnicholson (approved):
LGTM!

PR #106, opened 2022-10-14
Use absolute binary path in .desktop

wjt:
Previously, the Exec line would be set to a relative path like:

```ini
Exec=libexec/gnome-image-installer
```

This will, of course, fail. Set an absolute path here.

While we're here, remove an unused variable.

https://phabricator.endlessm.com/T33912


pwithnall (approved):
LGTM

PR #107, opened 2022-12-08
Fix missing format specifier in Turkish translation

dbnicholson:
Recent `msgfmt --check` fails since this translation is missing a format specifier. According to Google Translate, the updated translation is `The location you selected is insufficient, you need %s to reformat with %s.`.

https://phabricator.endlessm.com/T33861

### DIFF
--- a/po/tr.po
+++ b/po/tr.po
@@ -202,7 +202,7 @@ msgstr "Geçerli sürümü kullan:"
 #, no-c-format
 msgid ""
 "The location you have chosen is too small: you need %s to reformat with %s."
-msgstr "Seçtiğiniz konum yetersiz, ihtiyacınız olan %s"
+msgstr "Seçtiğiniz konum yetersiz, ihtiyacınız olan %s ile yeniden formatlamak için %s."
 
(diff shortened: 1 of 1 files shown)


po/tr.po
```diff
@@ -202,7 +202,7 @@ msgstr "Geçerli sürümü kullan:"
 #, no-c-format
 msgid ""
 "The location you have chosen is too small: you need %s to reformat with %s."
-msgstr "Seçtiğiniz konum yetersiz, ihtiyacınız olan %s"
+msgstr "Seçtiğiniz konum yetersiz, ihtiyacınız olan %s ile yeniden formatlamak için %s."
```
wjt, line 205: But nothing will push this into Transifex, so the next pull will overwrite this with the bad string. Do you have a Transifex account? If so you can edit the Turkish translation there. If not, I can make the change there.

wjt:
I copied your translation into Transifex.

po/tr.po
dbnicholson: I do have an account. That's how I've been doing all the rebasing. Good to know about how the process goes, though.

wjt: Of course. Sorry, slow brain day!

PR #109, opened 2023-04-09
Tell Meson to use native programs where appropriate

alyssais:
These programs are used as part of the build, so Meson should use their native versions.

As a further enhancement, it would be nice if the tests could be disabled (e.g. when cross compiling to an architecture we can't run the binaries for), possibly with a Meson option.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -15,20 +15,20 @@ test_env.set('GIO_MODULE_DIR', '')
 
 # TODO: it must surely be possible to do better than this with meson
 # generators
-gpg = find_program('gpg2', 'gpg')
-gpgconf = find_program('gpgconf')
+gpg = find_program('gpg2', 'gpg', native : true)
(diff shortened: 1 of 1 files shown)


dbnicholson:
ok to test

dbnicholson (approved):
Makes sense. And it would be nice to disable the tests when they're not found. I can't think of a nice way to do that at the moment, though.

eli-schwartz:
> Makes sense. And it would be nice to disable the tests when they're not found. I can't think of a nice way to do that at the moment, though.

You can try adding `required: false, disabler: true` to the find_program lookups. This will mean:
- it's allowable to not find them
- if they are not found, then the variable becomes a disabler object, which is a "contagious" no-op and causes all language constructs where it is referenced (for example, as parameters to the `test()` function) to also become disabler objects. End result: those tests are not run.

PR #110, opened 2024-01-16
icon: Refresh with scalable icon in new GNOME style

cassidyjames:
I've been hacking on a GNOME-style icon for something else in my free time, so I just recently became more comfortable with the icon creation process. I've thrown this together using the GNOME guidelines. The logo is, well, our logo, and the down arrow is adapted from the GNOME Icon Library, licensed as CC0 (public domain).

I could probably clean up the Meson file a bit more, but this _should_ be the minimally-invasive change to get the new icon.

https://phabricator.endlessm.com/T35115


manuq (approved):
:+1: this is much better than the previous pseudo-3D version.

PR #112, opened 2024-03-28
finished: Use GNOME Text Editor if present

wjt:
We would like to switch from preinstalling gedit to preinstalling GNOME Text Editor. Check if Text Editor is installed and use it if so; falling back to Gedit.

(For extra credit, we could use
`g_app_info_get_all_for_type ("text/plain")`, but we do want to hardcode a good default because, as explained in the comment, in the Initial Setup session all mime types are associated with `/bin/true` to prevent launching external applications, so we can't rely on `g_app_info_get_default_for_type ("text/plain")` returning something useful.)

https://phabricator.endlessm.com/T32962

### DIFF
--- a/gnome-image-installer/pages/finished/gis-finished-page.c
+++ b/gnome-image-installer/pages/finished/gis-finished-page.c
@@ -47,7 +47,7 @@
 struct _GisFinishedPagePrivate {
   GtkAccelGroup *accel_group;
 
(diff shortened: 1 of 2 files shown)


dylanmccall (approved):
This builds and runs for me, and it looks sane. No complaints :+1:

PR #114, opened 2024-04-18
Mount iso9660 partitions read-only

dsd:
On live boots where we now boot directly from loopback (without having an additional device-mapper layer), we need to mount (for a 2nd time) the image partition in order to find the data needed by the installer.

For some reason, the mount command fails to handle this case, because it (per its default) tries a rw mount and gets EBUSY from the kernel, then it gives up.

If you try to double-mount an iso9660 partition in a simple test you get different behaviour: for both mounts, it attempts rw first as normal, the kernel returns EACCES and mount falls back to 'ro'.

Not exactly sure why the behaviour is different in this context, however forcing read-only mounts for iso9660 does avoid the issue.

https://phabricator.endlessm.com/T35126

### DIFF
--- a/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
+++ b/gnome-image-installer/pages/diskimage/gis-diskimage-page.c
@@ -654,8 +654,24 @@ gis_diskimage_page_mount (GisPage *page)
       else
         {
           g_autofree gchar *path = NULL;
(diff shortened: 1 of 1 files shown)


dsd:
Am I leaking `options` here? (same as the original code before my modification?)

dbnicholson (approved):
Makes sense.

PR #115, opened 2024-04-26
disktarget: Skip loop devices which are mounted filesystems

wjt:
In particular, this will prevent the loop device that is bound to
persistent.img on persistent live media from being offered as an install
target.

https://phabricator.endlessm.com/T35347


dbnicholson (approved):
Yeah, that's a good idea in general. Wish I'd thought of that at the time.

PR #117, opened 2024-08-19
Miscellaneous cleanup

wjt:
While looking at this project I found various warning fixes lying around in poorly-named branches and commits.


dbnicholson (approved):
All seems reasonable to me.

